PR #3, opened 2018-03-02
Move to promises instead of promisify (for node v6 support)

alex-in2:


package.json
```diff
@@ -39,6 +39,7 @@
     "ajv-errors": "^1.0.0",
     "colors": "^1.1.2",
     "recursive-readdir": "^2.2.1",
-    "semver": "^5.5.0"
+    "semver": "^5.5.0",
+    "util.promisify": "^1.0.0"
```
wwsno: It's not used anywhere in the package?

PR #6, opened 2018-04-13
Feature/csh 5299 complete schema

vitaly-retel:


lib/components-schema-v1_0_x.ts
```diff
@@ -95,29 +99,57 @@ export const componentsDefinitionSchema_v1_0_x = {
                                     options: {
                                         type: 'array',
                                         minItems: 1,
+                                        items: {
+                                            type: 'object',
+                                            properties: {
+                                                caption: {
```
wwsno: Can you add descriptions for the new properties added?

vitaly-retel: sure

lib/components-types-v1_0_x.ts
```diff
@@ -83,36 +83,62 @@ export interface ComponentsDefinitionV10X {
...
+          type: "select";
+          options: {
+            /**
+             * Label of the item
+             */
+            caption: string;
+            /**
+             * Value of the item
+             */
+            value?: string;
```
wwjhu: Maybe good to add to the comment why you can leave the value out.

lib/components-types-v1_0_x.ts
```diff
@@ -83,36 +83,62 @@ export interface ComponentsDefinitionV10X {
...
+           */
+          pattern?: string;
+          /**
+           * Default value which is used instead of empty value
+           */
+          defaultValue?: string;
+          /**
+           * Unit type
+           */
+          unit?: string;
```
wwjhu: Add to the comment examples of units

lib/components-types-v1_0_x.ts
```diff
@@ -83,36 +83,62 @@ export interface ComponentsDefinitionV10X {
...
           [k: string]: any;
         }
       | {
           /**
            * Color picker field property
            */
-          type?: "colorPicker";
-          [k: string]: any;
+          type: "colorPicker";
+          options?: {
```
wwjhu: Why is the opacity in an options object while for other fields properties are siblings?

lib/components-types-v1_0_x.ts
wwsno: Do keep in mind this file is generated from lib/components-schema-v1_0_x.ts, so the changes should go in there.

lib/components-types-v1_0_x.ts
wwsno: Guess we could drop the options thing?

lib/components-schema-v1_0_x.ts
```diff
@@ -95,29 +99,63 @@ export const componentsDefinitionSchema_v1_0_x = {
...
                                     },
+                                    pattern: {
+                                        type: 'string',
+                                        description: 'Value validation regexp pattern',
+                                    },
+                                    defaultValue: {
+                                        type: 'string',
+                                        description: 'Default value which is used instead of empty value',
+                                    },
+                                    unit: {
```
wwsno: Just wondering, is it a limited number of string options we support here? If so we could further restrict it?

PR #8, opened 2018-04-20
CSH-5296: Add radio control type

vitaly-retel:


lib/components-schema-v1_0_x.ts
```diff
@@ -117,6 +117,39 @@ export const componentsDefinitionSchema_v1_0_x = {
...
+                                        items: {
+                                            type: 'object',
+                                            additionalProperties: false,
+                                            required: ['caption', 'icon'],
+                                            properties: {
+                                                caption: {
+                                                    type: 'string',
+                                                    description: 'Label of the item',
+                                                },
+                                                icon: {
```
wwsno: Should we do something on the backend for the icon? Similar to the way component icons are converted to urls?

vitaly-retel: Definitely

PR #10, opened 2018-04-25
Add IDE and editors to .gitignore

wwjhu:


vitaly-retel:
why did you add 'tmp' folder, does one of IDE use it?

wwjhu:
I added it as it is a common folder to ignore like `dist`. I can remove it again if you think it is better.

PR #11, opened 2018-04-25
CSH-5293: Add disable-fullscreen-checkbox property

vitaly-retel:


wwjhu:
My impression from earlier discussions was that we would not turn the fullscreen property into a component property but make it hardcoded in the editor.

vitaly-retel:
Oh really? We should have added a comment into the description then as we did for parallax property. What to do now?

wwsno:
Not sure what we discussed. Might make sense. Would it be a lot of work to change?

PR #12, opened 2018-04-27
CSH-5295: Add drop-capital control type

vitaly-retel:


lib/components-schema-v1_0_x.ts
```diff
@@ -224,6 +224,16 @@ export const componentsDefinitionSchema_v1_0_x = {
                                     },
                                 },
                             },
+                            {
```
wwjhu: Looking at what else is here, I am really wondering whether dropcap should be a separate type. To me it is a property composed of three text fields instead.

vitaly-retel: But they all should always be together. And there is a special UI for them too.

wwjhu: What's special about that UI? I see three number fields that do are not interconnected, right? 

Maybe I am wrong here but I would expect the `drop-capital` componentProperty to consists of 3 controls instead.

PR #13, opened 2018-05-08
Feature/csh 5294 social prop validation

wwsno:


lib/components-schema-v1_0_x.ts
```diff
@@ -196,6 +196,10 @@ export const componentsDefinitionSchema_v1_0_x = {
                                         type: 'string',
                                         description: 'Input placeholder',
                                     },
+                                    readonly: {
+                                        type: 'string',
```
wwjhu: Intentionally string and not boolean?

vitaly-retel: yeah, seems should be boolean?

wwsno: Fixed.

PR #15, opened 2018-05-18
fitting-property CSH-3375: introduce control type for fitting propert…

wwsno:
…y, as it's not customizable.


vitaly-retel:
Types should be updated too right?

wwsno:
Updated types.

PR #16, opened 2018-05-29
CSH-3793: Support of interactive component

vitaly-retel:


lib/components-schema-v1_0_x.ts
```diff
@@ -293,6 +293,28 @@ export const componentsDefinitionSchema_v1_0_x = {
...
+                                properties: {
+                                    type: {
+                                        enum: ['interactive'],
+                                        description: 'Configuration of interactive directive',
+                                    },
+                                    defaultConfig: {
+                                        type: 'object',
+                                        description: 'Default configuration of interactive directive',
+                                    },
+                                    editLink: {
```
wwsno: Looks like it's possible to automatically validate urls:
https://spacetelescope.github.io/understanding-json-schema/reference/string.html#built-in-formats

vitaly-retel: yeah

vitaly-retel: not sure how strict it should be, I think general 'uri' should be enough

lib/components-schema-v1_0_x.ts
```diff
@@ -305,6 +329,7 @@ export const componentsDefinitionSchema_v1_0_x = {
                             'doc-html',
                             'doc-slideshow',
                             'doc-media',
+                            'doc-interactive',
                         ],
                     },
                     dataProperty: { type: 'string' },
```
vitaly-retel: guess this one can be removed too?

wwsno: Sure.

lib/components-schema-v1_0_x.ts
wwsno: Think so too.

PR #17, opened 2018-06-06
CSH-3293: Add base part (parser) for additional validators. Add restr…

vitaly-retel:
…ictChildren validator.


lib/validators/restrict-children-validator.ts
```diff
@@ -0,0 +1,69 @@
...
+            if (!isPresent) {
+                if (this.hasSlideshowDirective(parsedComponent)) {
+                    errorReporter(`Component property "${PROPERTY}" must be defined in component "${parsedComponent.component.name}" because the ` + 
+                        `component contains slideshow directive`);
+                    valid = false;
+                }
+                return;
+            }
+            // check if all keys point to corrent component
+            // TODO if || {} is removed then ES2017 shows an error that parsedComponent.component[PROPERTY] may be undefined, why?
```
vitaly-retel: Does anyone have an idea?

If strict mode is off (tsconfig) then it builds. Otherwise need to add that '|| {}'. With ES2015 it builds too.

test/parser/parser-util.test.ts
```diff
@@ -0,0 +1,252 @@
...
+                return Promise.resolve(result);
+            };
+
+            let er = '';
+            try {
+                await parseDefinition(componentsDefinition, getFileContent)
+            } catch(e) {
+                er = e.message;
+            }
+            expect(er).toEqual(`Directive's attributes must be uniq. Attribute value is "text"`);
```
wwsno: unique?

test/parser/parser-util.test.ts
```diff
@@ -0,0 +1,252 @@
...
+                                    'dataType': 'doc-image'
+                                },
+                                'directiveKey': 'image'
+                            }
+                        }
+                    }
+                }
+            };
+            const getFileContent = (filePath: string) : Promise<string> => {
+                const filename = filePath.split(/[\/\\\\]/).pop();
```
wwsno: It's nicer to use the path module.

```
import * as path from 'path';
path.basename(filePath);
```

lib/models.ts
```diff
@@ -0,0 +1,35 @@
...
+    media = 'media',
+    interactive = 'interactive',
+
+    unknown = 'unknown',
+}
+
+export interface ParsedComponentsDefinition {
+    components: {
+        [name: string]: {
+            component: ComponentsDefinitionV10X['components'][0];
```
wwsno: I wonder how this will work out for future versions. At some point we might have a `ComponentsDefinitionV11X`. We might get something like `ComponentsDefinitionV10X['components'][0] | ComponentsDefinitionV11X['components'][0]` here.

lib/parser/parser-utils.ts
```diff
@@ -0,0 +1,94 @@
...
+import { DirectiveType, ParsedComponentsDefinition } from '../models';
+
+/**
+ * Returns information about directives
+ *
+ * @param content HTML content of the component
+ * @returns ParsedComponentsDefinition['components']['name']['directives']
+ */
+function parseDirectives(content: string) : ParsedComponentsDefinition['components']['name']['directives'] {
+    const regexTags = /<([a-z_\-]+).*?\s+doc-[a-z\d\-]+=['"]?[a-z\d\-]+['"]?.*?>/gmi;
```
wwsno: I wouldn't use regex to parse the html content, but rather an html parser like htmlparser2 and walk through the attributes.

lib/validators/restrict-children-validator.ts
```diff
@@ -0,0 +1,69 @@
...
+            const propertyValue = parsedComponent.component[PROPERTY] || {};
```
wwsno: What if restrictChildren object is left empty? It would still pass.

Also, when a doc-slideshow directive is present, I would limit it to exactly one entry (as multiple entries don't really make sense).

lib/validators/restrict-children-validator.ts
wwsno: Guess Typescript can't infer that you already checked if the property exists?

Apparently there is an operator to assert that the value is non null (Non-null assertion operator):

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "ES2015",
+    "target": "ES2017",
```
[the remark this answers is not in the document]
wwsno: You could opt for only adding the library definition with Object.values and still target ES2015.

Node 8 does support Object.values: https://node.green/

lib/validators/restrict-children-validator.ts
wwjhu: I think Sander is right about TS not being able infer. Adding an `if` solves it (very ugly though). Upgrading to TS 2.9.1 didn't help either

lib/models.ts
wwjhu: Besides that, it does not become more readable with types like `ComponentsDefinitionV10X['components'][0]` and `ParsedComponentsDefinition['components']['name']`. That would be something that we could address in `generateComponentTypes.ts`, right?

test/parser/parser-util.test.ts
vitaly-retel: done

lib/models.ts
vitaly-retel: Added generic types. Probably it solves it partly at least.

lib/parser/parser-utils.ts
vitaly-retel: ok

lib/validators/restrict-children-validator.ts
vitaly-retel: Updated the schema

tsconfig.json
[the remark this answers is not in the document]
vitaly-retel: Updated the config, is it ok now?

lib/validators/components-validator.ts
```diff
@@ -0,0 +1,82 @@
...
+                valid = false;
+                errorReporter(`Component "${comp.name}" style scss file missing "${componentStylePath}"`);
+            }
+        }
+
+        // generic files
+        for (const file of GENERIC_FILES) {
+            if (!this.filePaths.has(file)) {
+                valid = false;
+                errorReporter(`Generic file "${file}" is missing`);
```
wwsno: I think you can just say "File ${file} is missing". It being "generic" might not say much to the integrator?

lib/validators/unit-type-validator.ts
```diff
@@ -0,0 +1,36 @@
...
+    ) {
+    }
+
+    validate(
+        errorReporter: (errorMessage: string) => void,
+    ): boolean {
+        let valid = true;
+
+        for (const property of this.definition.componentProperties) {
+            if (property.control.type === 'text' && property.control.unit && !TYPES_REGEXP.test(property.control.unit)) {
```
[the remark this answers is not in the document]
vitaly-retel: it does now

lib/validators/components-validator.ts
vitaly-retel: ok

PR #18, opened 2018-06-08
Feature/csh 3954 add slides property

wwsno:
* New control types slides and header.
* New option for properties list in components to specify as an object, same as in componentProperties. Allows merging them together.


lib/components-schema-v1_0_x.ts
```diff
@@ -40,7 +336,18 @@ export const componentsDefinitionSchema_v1_0_x = {
...
                         type: 'array',
-                        items: { type: 'string' },
+                        items: {
+                            oneOf: [
+                                {
+                                    type: 'string',
+                                },
+                                {
+                                    type: 'object',
+                                    additionalProperties: false,
```
vitaly-retel: seems you forgot about required: ['name', 'label', 'control', 'dataType'] ?

lib/components-schema-v1_0_x.ts
```diff
@@ -5,6 +5,302 @@
...
  /* tslint:disable:max-line-length variable-name */
+
+const componentPropertyDefinition = {
+    name: {
+        type: 'string',
+        description: 'Unique identifier of component property',
+        minLength: 3,
+    },
+    label: { type: 'string', description: 'Display label of Component property' },
+    directiveKey: { type: 'string', description: 'Directive key for properties that use a directive data type' },
```
vitaly-retel: It looks differently that an example which you showed in slack. If I understand correctly the whole property configuration should be described if directiveKey is set? For example I have two components, both have image directive but directives key are different. Now how to apply image-editor property to them?

vitaly-retel: Iteresting that changes in boilerplate project is the same as an example in slack. So why not to describe all three posible variants of property item of properties field? Now there are two only and the second one kind of covers anonymous property and reference property with directiveKey.

lib/components-schema-v1_0_x.ts
```diff
@@ -5,6 +5,302 @@
...
+                        description: 'Enables fitting option for an image directive',
+                    },
+                },
+            },
+            {
+                additionalProperties: false,
+                required: ['type'],
+                properties: {
+                    type: {
+                        enum: ['slides'],
```
vitaly-retel: Seems this control type requires additional checkings in the validator like
- directive key should be valid
- works with doc-slideshow only
- other?
Can you please add a bullet to my current ticket?

lib/components-schema-v1_0_x.ts
wwsno: except none are required. Whether it becomes a valid object depends. If name is set, it will be merged with the matching property in componentProperties.

This might end up requiring some additional validation.

lib/components-schema-v1_0_x.ts
wwsno: I will add it to the current pull request after you merge your ticket to master.

lib/components-schema-v1_0_x.ts
wwsno: Not sure what you mean, but the entry from componentProperties is merged with the object from the properties array of a component property. The directiveKey can be specified in the properties list of the component.

vitaly-retel: I mean that now it is possible to add non valid property, for example without name and directiveKey, or other variant. Now the schema does not guarantee that the item added to 'properties' property of component is valid

wwsno: Name is not required.

I will see if I can add validation after you merge your task.

lib/models.ts
```diff
@@ -26,12 +26,7 @@ export interface ParsedComponentsDefinition {
                     tag: string;    // tag name (lowercased)
                 };
             };
-            properties: {
-                [name: string]: {
-                    property: ComponentsDefinition['componentProperties'][0];
-                    directiveKey: string | null;
-                };
-            };
+            properties: ComponentsDefinition['componentProperties'][0][];
```
wwjhu: Should we add a type for `ComponentsDefinition['componentProperties'][0]`?

wwsno: This one is the autogenerated type from the schema. You can grab the nested type like this.

wwjhu: It's just reads a bit awkward

lib/index.ts
```diff
@@ -120,6 +120,7 @@ export async function validate(
         new PropertiesValidator(filePaths, componentsDefinition),
         new RestrictChildrenValidator(parsedDefinition),
         new UnitTypeValidator(componentsDefinition),
+        new SlidesValidator(parsedDefinition),
```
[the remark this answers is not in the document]
wwsno: Done.

lib/validators/slides-validator.ts
```diff
@@ -0,0 +1,107 @@
...
+        errorReporter: (errorMessage: string) => void,
+        parsedComponent: ParsedComponentsDefinitionComponent,
+        property: ParsedComponentsDefinitionProperty,
+    ): boolean {
+        if (property.control.type !== 'slides') {
+            return true;
+        }
+
+        if (!parsedComponent.component.restrictChildren) {
+            errorReporter(`Component "${parsedComponent.component.name}" must have restrictChildren set to use the slides property`);
```
vitaly-retel: Hmm, there is restrict-children-validator as well. Now the whole validator checks if restrictChildren is present for doc-slideshow directive and for slides control. Probably it can be combained somehow. But it could be like it is now too I think

wwsno: I will see if I can pick it up in between, there's some other small things I would like to refactor as well (in particular properties list being in two places). However will merge it now to progress on the slideshow properties.

PR #19, opened 2018-06-13
CSH-3376: Make css class of disable fullscreen property hardcoded

vitaly-retel:


lib/components-schema-v1_0_x.ts
```diff
@@ -164,13 +164,13 @@ export const componentsDefinitionSchema_v1_0_x = {
                             },
                             {
                                 additionalProperties: false,
-                                required: ['type', 'value'],
+                                required: ['type'],
                                 properties: {
                                     type: {
                                         enum: ['disable-fullscreen-checkbox'],
-                                        description: 'Checkbox toggling between value and no value with additional checking of link directives. The property is set and disabled if the component has a non empty link directive',
+                                        description: 'Checkbox toggling between value and no value with additional checking of link directives. ' +
```
wwjhu: Why switch to multi-line here?

PR #21, opened 2018-06-21
CSH-3722 - Make all label fields accept an object with string key and optional values next to a plain string

wwjhu:


test/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
```
vitaly-retel: I'm not sure but maybe instead of having completely different tsconfig.json we can use extending https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#configuration-inheritance-with-extends ?

wwsno: I also wonder if it's really needed?

wwjhu: The TS error checking in the IDE searches up the directory tree for the nearest `tsconfig.json`. It will find the one in the root, but that has an `include` that does obviously does not contain the tests. As a result the TS compiler reports a "Cannot find parent 'tsconfig.json'" error and doesn't show errors for the tests. The solution is to have the tests have their own tsconfig.json which includes just the tests. It's useful functionality to have: we currently have 50+ TS errors in the tests. Probably all minor, but nonetheless useful to fix.

I can modify the tsconfig to derive the properties from the parent, but we can also opt to remove most of the options.
Deriving will look like
```
{
  "extends": "../tsconfig",
  "include": [
      "./index.test.ts",
      "./**/*"
  ]
}
```

PR #23, opened 2018-07-03
validator-improvements CSH-5678: Add several validator checks on icon…

kgeisink:
…s, doc-media and properties.

Check if icons have SVG or PNG format; 
check if PNG files are transparent; 
check that a component has only one doc-media directive; 
check media-properties control type for doc-media properties


lib/validators/properties-validator.ts
```diff
@@ -50,6 +51,14 @@ export class PropertiesValidator implements Validator {
                     }
                 }
             }
+
+            // Validate if the doc-media properties have a media-properties control type
+            if (compProp.dataType === 'doc-media') {
```
vitaly-retel: This checking doesn't look correct and it is not full. The goal is to check if a component uses doc-media directive and if it does then check if the component has a property with control type media-properties and this property is applied to that directive (checking directiveKey). I was going to suggest you looking at interactive directive and property checking but it is not implemented too. At least you can take a look how property type is checked in interactive-validator, because the current one is wrong, if dataType is not doc-media then control type can be 'media-properties' - what is wrong. You have to check control type first, then dataType.

lib/validators/icons-validator.ts
```diff
@@ -0,0 +1,44 @@
...
+    validate(
+        errorReporter: (errorMessage: string) => void,
+    ): boolean {
+        let valid = true;
+
+        Object.values(this.definition.components).forEach((component) => {
+            const ext = path.extname(component.icon).toLowerCase();
+            if (this.supportedFormats.indexOf(ext) === -1) {
+                errorReporter(`Icons are only supported in SVG or transparent PNG format`);
+                valid = false;
```
vitaly-retel: Should we denied other formats at all, or just show a warning message that it may look bad in the end?

wwsno: We could start with this and wait for feedback from customers?

lib/validators/properties-validator.ts
wwsno: Might indeed make sense to move it to DocMediaValidator:
- When the component has more than 1 doc-media, the component is invalid
- When it has 1 doc-media directive, it can have a property with dataType='doc-media' + control.type='media-properties'
- When it has no doc-media directive, it can't have a property with control.type='media-properties'.

kgeisink: @vitaly-retel The jira task this was based on was by far not as explanatory of the required behaviour as that. I'll look into it.
@wwsno Thanks for the summary of how it should behave, I can work with this.

lib/validators/icons-validator.ts
kgeisink: From what I understood it was decided a while ago to explicitly deny other file formats. This was also OK'd by Tom etc. (It's also described as such in the jira issue.)

package.json
```diff
@@ -43,8 +43,10 @@
     "ajv": "^6.1.1",
     "ajv-errors": "^1.0.0",
     "colors": "^1.2.1",
+    "fs": "^3.9.2",
```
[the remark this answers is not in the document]
kgeisink: I found a list now that mentions which packages are included!

lib/validators/properties-validator.ts
vitaly-retel: PropertiesValidator checks properties list only. But everything what is related to the directives should check components. You can see that there is ComponentsDefinition and ParsedComponentsDefinition. The last one includes information about directives: which directives components have and which properties those directives use. As I said PropertiesValidator checks ComponentsDefinition (simple checking if properties list is ok). Better to place them all in DocMediaValidator. And again, the same should be done for interactive directive. Guess you can combine it somehowe because validation logic is the same for both doc-media and doc-interactive.

kgeisink:
How's this?

lib/validators/doc-media-validator.ts
```diff
@@ -1,9 +1,15 @@
 /**
  * Validates usage of doc-media directive
+ *
+ * Rules:
+ *  - A component is not allowed to have more than one doc-media directive.
+ *  - A component with 1 doc-media directives -can- have a property control type "media-properties"
```
vitaly-retel: I think if a component has doc-media directive then it must have a property control type media-properties. Right, Sander?

lib/validators/doc-media-validator.ts
```diff
@@ -24,9 +30,26 @@ export class DocMediaValidator implements Validator {
...
+                let mediaProperties = Object.values(parsedComponent.properties).filter((property) => property.control.type === 'media-properties');
+                // Check whether the component has a media-properties control type
+                if (numMediaDirectives < 1 && mediaProperties.length > 0) {
+                    errorReporter(`Only components with a doc-media directive can have a "media-properties" control type`);
+                    valid = false;
+                }
+                if (numMediaDirectives === 1) {
+                    // Check whether the media-properties control type property is applied to the doc-media directive.
+                    Object.values(mediaProperties).forEach((mediaProperty) => {
+                        if (mediaProperty.directiveKey !== 'media') {
```
vitaly-retel: Directive key can be any value. So you may check for exising of the media directive like
const directive = parsedComponent.directives[mediaProperty.directiveKey];
if (!directive || directive.type !== DirectiveType.media) {
  // property points to non exising directive or non doc-media directive
}

lib/validators/doc-media-validator.ts
wwsno: It's not required. It just means you don't have the dynamic social media properties in the sidebar.

vitaly-retel: Ah right, sorry. It's the interactive directive requires the property.

PR #31, opened 2018-08-15
CSH-5824: Supporting of autofill component property

vitaly-retel:


lib/models.ts
```diff
@@ -16,27 +17,30 @@ export enum DirectiveType {
...
     groups: {
-        [name: string]: ComponentsDefinition['groups'][0];
+        [name: string]: T['groups'][0];
     };
     defaultComponentOnEnter: string;
-    conversionRules: ComponentsDefinition['conversionRules'];
+    conversionRules: T['conversionRules'];
 }
+export type ParsedComponentsDefinitionV10X = ParsedComponentsDefinition<ComponentsDefinitionV10X>;
+export type ParsedComponentsDefinitionV11X = ParsedComponentsDefinition<ComponentsDefinitionV11X>;
```
vitaly-retel: quite shitty, not sure how to handle it nicely

wwsno: The type would become optional with a note in the comment that it's removed from version 1.x.x. The Digital Editor would need to support both cases anyway, since old component sets must always keep working. The types are only here to help us.

PR #32, opened 2018-08-16
CSH-5630: Override default component on enter property

vitaly-retel:


generateComponentsTypes.ts
```diff
@@ -19,3 +20,5 @@ compileSchema(componentsDefinitionSchema_v1_0_x,
     './lib/components-types-v1_0_x.ts', 'ComponentsDefinitionV10X');
 compileSchema(componentsDefinitionSchema_v1_1_x,
     './lib/components-types-v1_1_x.ts', 'ComponentsDefinitionV11X');
+compileSchema(componentsDefinitionSchema_v1_2_x,
```
wwsno: Please revert and merge changes into ComponentsDefinitionV11X. We only need to introduce a new version between new Content Station versions. The versions are tied to what the Content Station version supports.

vitaly-retel: yeah, right

PR #33, opened 2018-08-17
Feature/csh 5670 groups option doc container

wwsno:


lib/validators/doc-container-groups-validator.ts
```diff
@@ -0,0 +1,74 @@
...
+                continue;
+            }
+            if (!parsedComponent.directives[key]) {
+                valid = false;
+                errorReporter(`Component "${parsedComponent.component.name}" has a group for invalid directive "${key}"`);
+                continue;
+            }
+            if (parsedComponent.directives[key].type !== 'container') {
+                valid = false;
+                errorReporter(`Component "${parsedComponent.component.name}" has a group for directive "${key}" with incompatible type "${parsedComponent.directives[key].type}"`);
```
vitaly-retel: maybe also lets add that only 'container' is allowed directive?

wwsno: Done.

PR #35, opened 2018-08-28
tslint-and-type-fixes-tests: fix tslint and type errors in test folde…

wwsno:
…r. Update test command to also run tslint and tsc on the test folder.

### DIFF
--- a/tsconfig.test.json
+++ b/tsconfig.test.json
@@ -1,0 +1,6 @@
+{
+    "extends": "./tsconfig.json",
+    "include": [
+        "./test/**/*"
+    ]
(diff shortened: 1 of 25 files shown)


tsconfig.test.json
```diff
@@ -0,0 +1,6 @@
+{
```
wwsno: Separate tsconfig for test folder, as build command should only output results for lib folder to dist.

wwjhu: Not sure if it works fine this way with the inline checker in PHPStorm. Will check once in master.

PR #39, opened 2018-09-28
use-parsed- component-definition CSH-5854: use parsed component defin…

wwsno:
…ition in all validators, rather mixing the usage of the unparsed and parsed definition. Make writing new validators more consistent.

### DIFF
--- a/lib/models.ts
+++ b/lib/models.ts
@@ -33,8 +33,9 @@ export interface ParsedComponentsDefinition<T extends ComponentsDefinition> {
     groups: T['groups'];
     defaultComponentOnEnter: string;
     conversionRules: T['conversionRules'];
+    scripts: T['scripts'];
 }
-export type ParsedComponentsDefinitionV10X = ParsedComponentsDefinition<ComponentsDefinitionV10X>;
+export type ParsedComponentsDefinitionV10X = ParsedComponentsDefinition<ComponentsDefinitionV11X>;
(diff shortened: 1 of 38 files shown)


lib/models.ts
```diff
@@ -33,8 +33,9 @@ export interface ParsedComponentsDefinition<T extends ComponentsDefinition> {
     groups: T['groups'];
     defaultComponentOnEnter: string;
     conversionRules: T['conversionRules'];
+    scripts: T['scripts'];
 }
-export type ParsedComponentsDefinitionV10X = ParsedComponentsDefinition<ComponentsDefinitionV10X>;
+export type ParsedComponentsDefinitionV10X = ParsedComponentsDefinition<ComponentsDefinitionV11X>;
```
wwsno: Ugly, but will disappear in CSH-5856.

lib/validators/components-validator.ts
```diff
@@ -21,20 +21,19 @@ const GENERIC_FILES = [
 ];
 
 export class ComponentsValidator extends Validator {
-
     constructor(
         error: (errorMessage: string) => false,
-        private filePaths: Set<string>,
-        private definition: ComponentsDefinition,
+        definition: ParsedComponentsDefinitionV10X,
+        protected filePaths: Set<string>,
```
vitaly-retel: Can't be 'private' anymore?

wwsno: Fixed.

PR #44, opened 2018-12-27
SPA-1879: Extend component definition parsing so it can be used in th…

vitaly-retel:
…e digital-editor

- enable tslint rule eofline
- use package.json "files" property instead of npmignore file


vitaly-retel:
Component definition model is not correct now:
- component properties,  here can be anonymous property and property object where seems only "name" is required field
I didn't do it because it doesn't look trivial, better to do it as another task.

lib/parser/parser-utils.ts
```diff
@@ -150,3 +210,103 @@ function findComponentPropertyTemplate(
...
+    defaultComponentContent: ComponentSet['defaultComponentContent'],
+    componentName: string,
+    property: ComponentProperty,
+) : void {
+    // No default value
+    if (!property.defaultValue) {
+        return;
+    }
+
+    switch (property.dataType) {
```
wwsno: Can you make sure all cases have matching tests? Currently only styles is tested.

lib/parser/parser-utils.ts
```diff
@@ -5,8 +5,55 @@
...
+ * - if group names are unique
+ * - if components have HTML rendition
+ *
+ * @param componentsDefinition
+ * @param getFileContent
+ * @returns Promise<ParsedComponentsDefinition>
+ */
+export async function parseDefinition(
+    componentsDefinition: ComponentsDefinition,
+    getFileContent?: GetFileContentType,
```
wwsno: Does it need to be optional? Could we make it required?

vitaly-retel: But the componentsDefinition may contain the renditions, then getFileContent is not needed. Or what do you mean? Passing null or empty function?

test/parser/parser-util.test.ts
```diff
@@ -1,226 +1,268 @@
...
+                `<figure doc-image="image"></figure>` +
+                `<div>` +
+                    `<img doc-image="second-image">` +
+                    `<p doc-editable="caption"></p>` +
+                `</div>` +
+            `</div>`;
             getFileContent = (filePath: string) : Promise<string> => {
                 const filename = path.basename(filePath);
                 let result;
                 switch (filename) {
```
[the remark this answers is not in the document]
vitaly-retel: ok

lib/parser/parser-utils.ts
wwsno: Never mind, leave it for now I guess. The design of the components set api in the backend makes it inconsistent.

PR #48, opened 2019-06-05
chore: update dev dependencies

wwsno:


tslint.json
```diff
@@ -57,7 +57,6 @@
       "no-switch-case-fall-through": true,
       "no-trailing-whitespace": false,
       "no-unused-expression": true,
-      "no-use-before-declare": true,
```
wwjhu: Why remove this?

wwsno: The rule is deprecated and already covered by typescript.

PR #50, opened 2019-07-11
Feature/dependencies update

vitaly-retel:
Fix dependencies security issue
- update lodash

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,13 +42,13 @@
   },
   "dependencies": {
     "@types/htmlparser2": "^3.7.31",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@woodwing/csde-components-validator",
-  "version": "1.6.0",
+  "version": "1.6.1",
```
wwsno: We don't really need to publish a new version for this.

vitaly-retel: yeah, right

PR #51, opened 2019-07-12
Use needed lodash functions only instead of the whole library

vitaly-retel:


package.json
```diff
@@ -41,16 +43,14 @@
     "typescript": "^3.5.1"
   },
   "dependencies": {
-    "@types/htmlparser2": "^3.7.31",
-    "@types/lodash": "^4.14.136",
```
vitaly-retel: I don't think we need it as we use the library with require

PR #61, opened 2020-11-27
Change validation check for doc-media directive to not allow the directive without a "media-properties" component property.

wwsno:
Without it does not work at all in the Digital Editor, so this check was wrong. Tested a social media component without "media-properties" in the Digital Editor, and it will fail because it cannot show the dialog. So this should never have been allowed.

Refactor code to remove all the nesting (improving readability) and include the component name in the error messages.


ananovak (approved):

ruudjungbacker (approved):
🎉

wwjhu (approved):
lgtm

PR #62, opened 2020-12-04
Enforce validate function to be async

jaapvanblaaderen:
This prevents confusion when listing validators and calling validate functions that seems synchronous but are in fact async.


wwsno (approved):
lgtm 💯

PR #64, opened 2020-12-07
Feature/ecs 368 limit component set

jaapvanblaaderen:
Adds limits for the component set and more specifically, the custom data folder. Component set limitations are based on current usage in production.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -46,9 +46,9 @@ import {
     DocContainerGroupsValidator,
     ConversionShortcutsValidator,
     LocalizationValidator,
+    PackageValidator,
 } from './validators';
 import { listFilesRelativeToFolder } from './util/files';
-import { PackageValidator } from './validators/package-validator';
 
 const ajv = new Ajv({ allErrors: true, jsonPointers: true, verbose: true });
 
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -2,9 +2,7 @@ import * as colors from 'colors/safe';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as semver from 'semver';
-
 import * as Ajv from 'ajv';
-import * as recursiveReadDir from 'recursive-readdir';
 
 import { componentsDefinitionSchema_v1_0_x } from './components-schema-v1_0_x';
 import { componentsDefinitionSchema_v1_1_x } from './components-schema-v1_1_x';
@@ -43,6 +41,8 @@ import {
     ConversionShortcutsValidator,
     LocalizationValidator,
 } from './validators';
+import { listFilesRelativeToFolder } from './util/files';
+import { PackageValidator } from './validators/package-validator';
 
 const ajv = new Ajv({ allErrors: true, jsonPointers: true, verbose: true });
 
@@ -65,15 +65,7 @@ export const readFile: GetFileContentType = (pathToFile: fs.PathLike, options?: 
  * @param folderPath Path to folder containing the components.
  */
 export async function validateFolder(folderPath: string): Promise<boolean> {
-    folderPath = path.normalize(folderPath);
-
-    // List files, make relative to input folder and normalize.
-    const files = new Set(
-        (await recursiveReadDir(folderPath)).map((p) =>
-            path.normalize(p).replace(new RegExp(`^${folderPath.replace(/\\/g, '\\\\')}(/|\\\\)?`), ''),
-        ),
-    );
-
+    const files = await listFilesRelativeToFolder(folderPath);
     return validate(
         files,
         async (filePath: string, options?: GetFileContentOptionsType) =>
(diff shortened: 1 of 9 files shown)


lib/validators/package-validator.ts
```diff
@@ -0,0 +1,89 @@
...
+        private filePaths: Set<string>,
+        private getFileContent: GetFileContentType,
+    ) {
+        super(error, definition);
+    }
+
+    async validate(): Promise<void> {
+        const files: FileInfo[] = [];
+        for (let filePath of Array.from(this.filePaths)) {
+            const file = (await this.getFileContent(filePath)) as Buffer | string;
```
wwsno, line 32: Should we implement a new function next to getFileContent to get the file size? Loading the content into memory is not the way to go I think. `fs` likely has a functions to get the file size statistics (`stats`) and in the backend we might be able to use the information from the zip file.

lib/util/files.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    // List files, make relative to input folder and normalize.
+    return new Set(
+        (await listFilesRecursively(folderPath)).map((p) =>
+            path.normalize(p).replace(new RegExp(`^${folderPath.replace(/\\/g, '\\\\')}(/|\\\\)?`), ''),
+        ),
+    );
+}
+
+export async function listFilesRecursively(folderPath: string): Promise<string[]> {
```
wwsno, line 15: Does not need to be exported.

lib/validators/package-validator.ts
```diff
@@ -0,0 +1,89 @@
...
+        const folderSize = this.calculateFolderSize(files);
+        if (folderSize > MAX_COMPONENT_SET_SIZE_MB) {
+            this.error(
+                `At ${folderSize}MB, the component set exceeds the total maximum size of ${MAX_COMPONENT_SET_SIZE_MB}MB.`,
+            );
+        }
+    }
+
+    validateCustomData(files: FileInfo[]): void {
+        const customDataFiles = files.filter((file) => file.path.startsWith('custom/'));
```
wwsno, line 61: Actually not sure what we do at the moment in the other places, but should we take case sensitivity into account?

lib/validators/package-validator.ts
```diff
@@ -0,0 +1,89 @@
+/**
+ * Validates if the component set package size
```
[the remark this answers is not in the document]
jaapvanblaaderen: Yes please ;-)

lib/validators/package-validator.ts
ruudjungbacker: I think we really should.

lib/index.ts
```diff
@@ -43,6 +41,8 @@ import {
     ConversionShortcutsValidator,
     LocalizationValidator,
 } from './validators';
+import { listFilesRelativeToFolder } from './util/files';
+import { PackageValidator } from './validators/package-validator';
```
ruudjungbacker, line 45: Is there a reason why this validator is imported separately?

ruudjungbacker (commented):
I really think the comment by Sander about the file sizes should be looked at. The are some other things but I think these are minor problems.

lib/util/files.ts
```diff
@@ -0,0 +1,27 @@
...
+    let fileList: string[] = [];
+    const files = await fs.promises.readdir(folderPath);
+    for (const file of files) {
+        const folder = path.join(folderPath, file);
+        if ((await fs.promises.stat(folder)).isDirectory()) {
+            fileList = [...fileList, ...(await listFilesRecursively(folder))];
```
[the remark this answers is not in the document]
jaapvanblaaderen: Done

lib/validators/package-validator.ts
```diff
@@ -0,0 +1,89 @@
...
+                path: filePath,
+                size: file.length,
+            });
+        }
+        this.validateComponentSetFileCount(files);
+        this.validateComponentSetSize(files);
+        this.validateCustomData(files);
+    }
+
+    validateComponentSetFileCount(files: FileInfo[]): void {
```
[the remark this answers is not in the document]
jaapvanblaaderen: Done

lib/validators/package-validator.ts
jaapvanblaaderen: Done

lib/index.ts
jaapvanblaaderen: No, fixed it :-)

lib/util/files.ts
jaapvanblaaderen: Done

lib/validators/package-validator.ts
jaapvanblaaderen: Done. Will make sure this works properly in the ecs-backend as well.

ruudjungbacker (approved):
lgtm

wwsno (approved):

PR #65, opened 2020-12-14
CSH-7816: create new component definition schema for showToolbars option

wwsno:
This option will allow to always show the toolbar of components. The default is to only show on hovering the mouse on the component.


lib/components-schema-v1_5_x.ts
```diff
@@ -0,0 +1,596 @@
...
+                        description: 'names of properties this component can use',
+                    },
+
+                    selectionMethod: {
+                        enum: ['default', 'handle'],
+                        description:
+                            'How this component is selectable, by default the user can select inside the component to select it',
+                    },
+
+                    showToolbar: {
```
wwsno, line 431: This is the new option in the 1.5 version schema.

lib/models/components-definition.ts
```diff
@@ -79,6 +79,13 @@ export interface Component {
      * How this component is selectable, by default the user can select inside the component to select it
      */
     selectionMethod?: 'default' | 'handle';
+
+    /**
+     * When to show the toolbar. By default only show on hovering the mouse on the component.
+     * Use the "always" option to show the toolbar in all cases.
+     */
+    showToolbar?: 'default' | 'always';
```
wwsno, line 87: This is type annotation for usage in our Digital Editor code.

test/index.test.ts
```diff
@@ -44,4 +48,14 @@ describe('getValidators', () => {
...
     });
+
+    it('should return amount of validators for version >= 1.4.0', () => {
+        const validators = getValidators('1.4.0', <any>null, <any>null, <any>null, <any>null, <any>null);
+        expect(validators && validators.length).toEqual(26);
+    });
+
+    it('should return amount of validators for version >= 1.5.0', () => {
+        const validators = getValidators('1.5.0', <any>null, <any>null, <any>null, <any>null, <any>null);
+        expect(validators && validators.length).toEqual(26);
```
wwsno, line 59: There is one validator less because in 1.4 we dropped a validator related to a specific option of Adobe AEM.

wwjhu (approved):
👍

PR #68, opened 2021-01-11
ECS-368: Extract top level package validation in a static method

jaapvanblaaderen:
Static validateSize method is to be used in ecs-backed directly to validate the size of the zip file before unpacking it.


wwsno (approved):

ruudjungbacker (approved):
LGTM

PR #72, opened 2021-03-01
ECS-388: validate only properties with names

ananovak:


ruudjungbacker (approved):
LGTM

lib/validators/properties-validator.ts
```diff
@@ -41,11 +41,13 @@ export class PropertiesValidator extends Validator {
     private validateProperty(property: ComponentProperty, componentPropertyNames: Set<string>) {
         this.validateReservedPropertyName(property);
 
-        // Validate we have not seen the name yet
-        if (componentPropertyNames.has(property.name)) {
-            this.error(`Component property "${property.name}" is not unique`);
+        if (property.name) {
```
wwsno, line 44: Would move this into a different function so you can add a guard clause of the `property.name` check and flatten the code.

wwsno (approved):

PR #73, opened 2021-03-02
Feature/rename project

ruudjungbacker:
I propose that we rename this repo/library to `studio-component-set-tools` so we can add more generic tools to the library besides the validation. 

As there is no real renaming solution for NPM we could deprecate the current `csde-components-validator` package on NPM as described here: https://stackoverflow.com/questions/28371669/renaming-a-published-npm-module

I've also reset the version to 1.0.0 as this is a new package. Although I'm not sure if that is something we want. 

When this PR is merged we should also update the repo name itself in GitHub. As far as I can see this should not have any impact as customers can still download the 'old' validator from NPM and builds are created manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
-# csde-components-validator
+# studio-component-set-tools
 
-Validation module for Content Station Digital Editor component sets.
+Tools module for Studio Digital Editor component sets.
 
 ## Usage
 
-The module provides two public api methods for validating a component set.
+The module provides tooling to develop component sets. It contains two public api methods for validating a component set.
 
 ### validateFolder
 
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Tools module for Studio Digital Editor component sets.
 
 ## Usage
 
-The module provides tooling to develop component sets. It contains two public api methods for validating a component set.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,10 +1,10 @@
-# csde-components-validator
+# studio-component-set-tools
 
-Validation module for Content Station Digital Editor component sets.
+Tools module for Studio Digital Editor component sets.
 
 ## Usage
 
-The module provides two public api methods for validating a component set.
+The module provides tooling to develop component sets. It contains two public api methods for validating a component set.
```
jaapvanblaaderen, line 7: It actually contains three public validation methods 😇  Might be good to rephrase this a bit with mentioning the number of methods.

jaapvanblaaderen (approved):
Looks good, one tiny remark.

README.md
ruudjungbacker: Updated it, please have another look.

wwsno: `It contains for example...` -> `For example, it contains...`?

ruudjungbacker: That is better indeed.

wwsno (approved):
👍

PR #74, opened 2021-03-02
Replaced tslint with eslint.

ruudjungbacker:
The used eslint config is the same as in the AN custom channel project.


package.json
```diff
@@ -36,13 +36,17 @@
...
     "prettier": "^2.2.1",
-    "rimraf": "^2.6.2",
-    "ts-jest": "^26.4.0",
-    "ts-node": "^8.2.0",
-    "tslint": "^5.17.0",
-    "typescript": "^4.1.0"
+    "rimraf": "^3.0.2",
+    "ts-jest": "^26.5.2",
+    "ts-node": "^9.1.1",
+    "typescript": "^4.2.2"
```
jaapvanblaaderen, line 49: I guess these library updates were required for eslint?

jaapvanblaaderen (approved):
Looks good, the readme still states 'TSLint' so probably want to change that.

package.json
ruudjungbacker: These last 4 weren't as far as I know. I could revert them if you want to?

ruudjungbacker:
The readme has been updated.

package.json
wwsno: I'm ok with it.

PR #75, opened 2021-03-02
Fixed the eslint errors in the lib folder.

ruudjungbacker:
For now this PR is based on the feature/eslint branch. It only contains the fixes for the issues in the lib folder. The fixes for the test folder will follow in a separate PR.

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -3,7 +3,7 @@ import * as path from 'path';
 import * as semver from 'semver';
 import ajv, { ValidateFunction } from 'ajv';
 import addFormats from 'ajv-formats';
-import colors from 'colors/safe';
+import * as colors from 'colors/safe';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore no typings for json-source-map
 import * as jsonMap from 'json-source-map';
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -4,8 +4,6 @@ import * as semver from 'semver';
 import ajv, { ValidateFunction } from 'ajv';
 import addFormats from 'ajv-formats';
 import * as colors from 'colors/safe';
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore no typings for json-source-map
 import * as jsonMap from 'json-source-map';
 
 import { componentsDefinitionSchema_v1_0_x } from './components-schema-v1_0_x';
@@ -17,11 +15,11 @@ import { componentsDefinitionSchema_v1_5_x } from './components-schema-v1_5_x';
 
 import { parseDefinition } from './parser';
 import {
-    ComponentsDefinition,
     ComponentSet,
     GetFileContentType,
     GetFileContentOptionsType,
     GetFileSize,
+    ComponentsDefinition,
 } from './models';
 import {
     Validator,
@@ -215,7 +213,7 @@ export function validatePackageSize(size: number, errorReporter: (errorMessage: 
 async function getComponentsDefinition(
     componentsDefinitionContent: string,
     errorReporter: (errorMessage: string) => void,
-): Promise<jsonMap.ParsedJsonSourceMap<ComponentSet>> {
+): Promise<jsonMap.ParsedJsonSourceMap<ComponentsDefinition>> {
     try {
         return jsonMap.parse(componentsDefinitionContent);
     } catch (e) {
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,9 +1,10 @@
-import * as colors from 'colors/safe';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as semver from 'semver';
 import ajv, { ValidateFunction } from 'ajv';
 import addFormats from 'ajv-formats';
+import colors from 'colors/safe';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore no typings for json-source-map
 import * as jsonMap from 'json-source-map';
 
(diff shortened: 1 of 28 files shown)


lib/index.ts
```diff
@@ -203,7 +215,10 @@ export function validatePackageSize(size: number, errorReporter: (errorMessage:
 async function getComponentsDefinition(
     componentsDefinitionContent: string,
     errorReporter: (errorMessage: string) => void,
-): Promise<{ data: ComponentsDefinition | null, pointers: any }> {
+): Promise<{
+    data: ComponentsDefinition | null;
+    pointers: { [key: string]: { [key: string]: { line: number; column: number; pos: number } } } | null;
```
ruudjungbacker, line 220: As there are no typings for json-source-map I created my own minimal version.

lib/parser/parser-utils.ts
```diff
@@ -4,7 +4,7 @@
 
 import * as path from 'path';
 import * as htmlparser from 'htmlparser2';
-const merge = require('lodash.merge');
+import merge from 'lodash.merge';
```
ruudjungbacker, line 7: For this change I had to enable `esModuleInterop` in the tsconfig.json file.

lib/models/get-file-content.ts
```diff
@@ -6,7 +6,7 @@ export type GetFileContentOptionsType = { encoding: string | null } | null;
 /**
  * Return string if encoding is string and Buffer otherwise.
  */
-export type GetFileContentType = (filePath: string, options?: GetFileContentOptionsType) => Promise<any>;
+export type GetFileContentType = (filePath: string, options?: GetFileContentOptionsType) => Promise<string | Buffer>;
```
ruudjungbacker, line 9: As `fs.readFile` is used both types can be returned. The main problem is that the code now assumes it is a string in some places and a Buffer in others. Not sure if this is the way to go, but as this is a 'public API' I don't know if we can change it.

lib/index.ts
```diff
@@ -1,9 +1,10 @@
-import * as colors from 'colors/safe';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as semver from 'semver';
 import ajv, { ValidateFunction } from 'ajv';
 import addFormats from 'ajv-formats';
+import colors from 'colors/safe';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore no typings for json-source-map
```
jaapvanblaaderen, line 8: Maybe replace this with our own type definition for this library?

Something like:
```
declare module 'json-source-map' {
    export function parse(json: string): Object;
}
```

Could for example be stored in `types/json-source-map` with a reference in the compilerOptions:
`"typeRoots": ["./node_modules/@types", "./types"]`

lib/index.ts
jaapvanblaaderen: Might be fixed by creating our own typings? See also my other comment.

lib/models/get-file-content.ts
jaapvanblaaderen: This looks like the correct change to me. This method is used to fetch both binary files and text based files from the component set, IMO not something we want to change at this point. Downside is indeed that you need to cast when using it.

jaapvanblaaderen (commented):
Few general remarks:
- The schema files still have this: `/* tslint:disable:max-line-length variable-name */`
- Running `npm run eslint-lib` returns 5 warnings.

lib/parser/parser-utils.ts
wwsno: Not sure if this is a good idea. It might cause all downstream libraries to also enable `esModuleInterop`.

You can remove that setting and instead do `import * as merge from 'lodash.merge';`

wwsno: See for example this comment: https://github.com/apollographql/apollo-server/issues/1977#issuecomment-683235089

lib/index.ts
```diff
@@ -134,11 +140,17 @@ export async function validate(
             const jsonLines = componentsDefinitionContent.split('\n');
 
             validateSchema.errors.forEach((error) => {
+                if (!componentsDefinitionSourcePointers) {
```
wwsno, line 143: This is not part of fixing eslint errors.

lib/index.ts
```diff
@@ -191,7 +203,7 @@ export async function validate(
  * @param size component set package size in bytes
  * @param errorReporter called when there is a validation error
  */
-export function validatePackageSize(size: number, errorReporter: (errorMessage: string) => void) {
+export function validatePackageSize(size: number, errorReporter: (errorMessage: string) => void): boolean {
```
wwsno, line 206: What's the eslint rule for the return types? I don't mind them being required for public APIs, but might be annoying if they are also required for private functions.

tsconfig.json
```diff
@@ -5,6 +5,7 @@
         "moduleResolution": "node",
         "sourceMap": true,
         "declaration": true,
+        "esModuleInterop": true,
```
wwsno, line 8: Should be removed, not needed.

lib/parser/parser-utils.ts
ruudjungbacker: I had to fix it a bit differently but `import merge = require('lodash.merge');` fixed the problem.

tsconfig.json
ruudjungbacker: I is removed.

lib/index.ts
ruudjungbacker: It has been added, please take a look.

lib/index.ts
ruudjungbacker: True, but since the pointers are now typed it the compiler complaints they could be `null`. Therefore added this check.

lib/index.ts
ruudjungbacker: In this case we use the default settings: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-function-return-type.md

If we want to tweak this we probably also want this for the other repos.

jaapvanblaaderen (approved):

PR #76, opened 2021-03-05
Feature/eslint test fixes

ruudjungbacker:
This PR is based on the feature/eslint-lib-fixes branch. It fixes all the eslint errors in the `./test` folder. Only for the `./test` folder the `@typescript-eslint/no-explicit-any` rule is turned off as the Component Set interfaces aren't entirely correct. So fixing this became time consuming. Another reason to disable this rule for tests is that partial test objects first need to be casted to unknown and then to the ComponentSet type for example. Not sure if this is something that we want.


wwsno (approved):
👍

jaapvanblaaderen (approved):
Nice 🚀

PR #77, opened 2021-03-12
Copied over the component-info and process-templates functionality from ECS backend.

ruudjungbacker:
I'm not exactly sure about the direction to take so I wanted to ask you guys to take a look at this draft PR. The main goal is to generate the component set info file while downloading the component set in the Apple News custom channel. 

This would look something like: 

```typescript
const zip = new AdmZip(path.resolve(__dirname, '../resources/minimal-component-set.zip'));
const definition = JSON.parse(zip.getEntry('components-definition.json').getData().toString());
await processTemplates(async (relativePath: string) => {
    const entry = zip.getEntry(relativePath);
    return entry ? entry.getData().toString() : undefined;
}, definition);
const info = processInfo(definition);
```

The current code in the ECS backend can be adapted to get the templates from S3 as they are now. 

At this moment I only copied over the parts I need at this moment. This makes it a bit scattered all over the place, but on the other hand if we don't need it we don't have to complicate things. 

Please let me know what you think. Then I can update the readme and add some extra tests.


lib/utils.ts
```diff
@@ -0,0 +1,2 @@
+export { processFields, processInfo } from './definition/component-info';
```
ruudjungbacker, line 1: Not sure about these exports. They are a 'public' API so this would make sense. Maybe we should rename `index.ts` to `validate.ts` to have separate API entry points?

wwsno (commented):
Direction looks fine to me 👍

jaapvanblaaderen (commented):
I lik the direction as well Ruud!

lib/utils.ts
ruudjungbacker: This has been done in #79

lib/models/component-set-fields.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * Combines information about components, includes label, iconUrl
+ * and component related fields
+ * label and iconUrl are extracted from component-definition.json file, while fields are loaded from html templates
+ * Example:
+ * {
+ *   body: {
```
wwsno, line 7: Example could use some updating, such as the label and iconUrl.

lib/definition/component-info.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Creates partial info definitions of the component set.
+ *
+ * This information is provided with custom channel publish messages. Integrators can use
+ * the info to interpret the given digital article without additional API requests.
+ */
+export function processInfo(componentsDefinition: ComponentsDefinition): ComponentSetInfo {
+    return {
+        components: componentsDefinition.components.reduce((result, component) => {
+            result[component.name] = {
+                fields: parseFields((<{ html: string }>component.renditions)[ComponentRendition.HTML]),
```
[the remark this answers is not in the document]
ruudjungbacker: Yes, I figured that out indeed. The main problem in the types right now is that a lot of fields are optional right now, including the `renditions`. As it is hard for me to figure out why this is the case I solved it this way. Also because we use this rule: https://github.com/typescript-eslint/typescript-eslint/blob/v4.16.1/packages/eslint-plugin/docs/rules/no-non-null-assertion.md I'm not exactly sure how to solve this.

lib/definition/component-info.ts
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Creates partial info definitions of the component set.
+ *
+ * This information is used by the digital to print conversion.
+ */
+export function processFields(componentsDefinition: ComponentsDefinition): ComponentSetFields {
+    return componentsDefinition.components.reduce((result, component) => {
+        result[component.name] = {
+            label: component.label as string,
+            iconUrl: component.icon,
```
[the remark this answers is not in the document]
ruudjungbacker: I see, I didn't noticed that indeed. Also looking at the other comments you have made I'm not sure anymore whether we want the `processFields` function in this library. The `processInfo` function is used to generate the information that is also included in the SNS message so this needs to be shared. 

The `processFields` function generated the information used for the digital to print conversion. I don't think we should ever make this format publicly available. So on one field maybe we should keep this function in the ECS backend only. On the other hand, since it is mostly the same could should we keep them together. Or should we only have a `processInfo` call in this library and post process this in the ECS backend to add the fields information?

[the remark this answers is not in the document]
wwsno: I think it could be removed and stay in the backend. The information is pretty specific for displaying information in the management console. We have the other format as a more general structure. Calling processInfo and then post process/transform into the fields structure sounds like something we should look into?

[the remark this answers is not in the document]
ruudjungbacker: Ok thanks, then I will take a look at that.

test/definition/process-templates.test.ts
```diff
@@ -0,0 +1,60 @@
...
+                    icon: 'icons/components/body.svg',
+                    label: {
+                        key: 'COMPONENT_BODY_LABEL',
+                    },
+                    name: 'body',
+                    properties: [],
+                    renditions: {
+                        html: undefined,
+                        psv: undefined,
+                    },
```
jaapvanblaaderen, line 55: Or `renditions: {}`?

README.md
```diff
@@ -63,6 +63,25 @@ parseDefinition(componentsDefinition).then((componentSet) => {
 });
 ```
 
+### processInfo
```
jaapvanblaaderen, line 66: The example could be improved I guess. I think you already need quite deep knowledge of the code to understand what is expected as parameters and what the functions return as result. Maybe it's an idea to describe the input and output for both?

jaapvanblaaderen (commented):
Code itself looks good to me, just left small remarks.

lib/definition/process-templates.ts
```diff
@@ -0,0 +1,52 @@
...
+    componentsDefinition: ComponentsDefinition,
+): Promise<void> {
+    const templates = loadTemplates(assetResolver, componentsDefinition);
+
+    for (const component of componentsDefinition.components) {
+        component.renditions = {};
+        for (const rendition of renditionTypes) {
+            component.renditions[rendition] = templates[rendition][component.name]
+                ? await templates[rendition][component.name]
+                : '';
```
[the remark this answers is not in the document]
wwjhu:
```
component.renditions[rendition] = (await templates[rendition][component.name]) ?? ''
```

lib/models/component-set-fields.ts
ruudjungbacker: I updated those fields now.

test/definition/process-templates.test.ts
ruudjungbacker: Actually they should be empty strings. so the previous test was wrong. I updated this.

README.md
```diff
@@ -63,6 +63,25 @@ parseDefinition(componentsDefinition).then((componentSet) => {
...
+
+An example of possible usage:
+
+```ts
+import { processTemplates, processInfo } from '@woodwing/studio-component-set-tools/dist/parser';
+
+const componentsDefinition = getComponentsDefinitionJson();
+// Add the rendition information
+await processTemplates(async (relativePath: string) => {
+    return await fs.promises.readFile(relativePath).toString(),
```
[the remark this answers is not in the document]
ruudjungbacker: Stupid mistake.

lib/definition/process-templates.ts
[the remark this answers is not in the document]
ruudjungbacker: And this actually fixed an issue.

README.md
ruudjungbacker: I updated the name and the structure. I left the `processTemplates` and `processInfo` function out of the documentation as these probably aren't used by integrators.

test/definition/process-templates.test.ts
wwsno: Is it the behavior of the original code in the ecs backend?

wwsno (approved):
Looks ok.

README.md
```diff
@@ -66,6 +66,52 @@ const componentsDefinition = getComponentsDefinitionJson();
 const componentSet = await parseDefinition(componentsDefinition);
 ```
 
+### generateComponentSetInfo
+
+To retrieve the component set information as provided when publishing an article from Studio, the `generateComponentSetInfo` function can be used.
+
```
jaapvanblaaderen, line 72: Suggestion:

The `generateComponentSetInfo` function can be used to generate a plain overview of components and component fields. This information is useful to interpret digital articles created for the given component set. The input for the function is information from the component set: component set info and rendition files (html, psv, etc).

The first parameter allows for specifying the components definition. As second parameter, a function is expected that returns the contents of a given component set file file as string.

jaapvanblaaderen (approved):
Good to have one overall function that's documented 👍 

Left a small text suggestion.

README.md
ruudjungbacker: Updated the readme. Also with an example about the restrict children property.

PR #78, opened 2021-03-15
ECS-390: add ComponentPropertyControlSlider to component property con…

ananovak:
I added ComponentPropertyControlSlider support here to be able to create a property definition in DE with type `slider:`
`import { ComponentProperty } from '@woodwing/csde-components-validator/dist/models';`

### DIFF
--- a/lib/models/component-property-controls.ts
+++ b/lib/models/component-property-controls.ts
@@ -220,22 +220,10 @@ export interface ComponentPropertyControlInteractive {
 
 export interface ComponentPropertyControlSlider {
     type: 'slider';
-
-    /**
-     * Minimum value
-     */
-    minValue: number;
-    /**
-     * Maximum value
-     */
-    maxValue: number;
-    /**
-     * Default value
-     */
+    value?: number;
+    minValue?: number;
+    maxValue?: number;
     defaultValue?: number;
-    /**
-     * Step
-     */
     step?: number;
 }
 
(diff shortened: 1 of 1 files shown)


lib/models/component-property-controls.ts
```diff
@@ -217,6 +218,15 @@ export interface ComponentPropertyControlInteractive {
     viewLink: string;
 }
 
+export interface ComponentPropertyControlSlider {
```
jaapvanblaaderen, line 221: Some other controls seem to have documentation per property, might be nice for this one as well?

jaapvanblaaderen (commented):
Code is ok I guess, not really sure if this contains all the required changes? I guess it should also be listed in a new components schema?

lib/models/component-property-controls.ts
ananovak: I do not, it 's a generic property. And it has quite obvious param names

ananovak: > Code is ok I guess, not really sure if this contains all the required changes? I guess it should also be listed in a new components schema?

I am not sure I have to do it all together, the minimum needed is a new property type to be able to use in in DE

Maybe new components schema could be added in a separate PR?

ananovak: Discussed a bit this property type with Sander https://github.com/WoodWing/contentstation/pull/2169#discussion_r595915134
And it's not clear for now should be `defaultValue` mandatory or optional. And what behavior is expected if value is `undefined`.
Looks like this topic needs to be discussed with UX team

PR #79, opened 2021-03-16
chore: Renamed the module index to validate.

ruudjungbacker:
Renamed the index module to validate so we have clear API entry points.


wwsno (approved):
Lgtm 🧇

PR #80, opened 2021-03-23
CSH-7936 — stripStylingOnPaste

wwjhu:
Adds generic directiveOptions validator and stripStylingOnPasteValidator


lib/validate.ts
```diff
@@ -269,6 +274,7 @@ export function getValidators(
     if (semver.satisfies(version, '>=1.0.0')) {
         validators = validators.concat(
             new ComponentsValidator(error, componentSet, filePaths),
+            new DirectiveOptionsValidator(error, componentSet),
```
wwjhu, line 277: Is it safe to add this validator to 1.0+?

test/resources/minimal-sample-1_6_x/components-definition.json
```diff
@@ -0,0 +1,97 @@
...
+        {
+            "name": "intro",
+            "label": {
+                "key": "INTRO_KEY"
+            },
+            "icon": "icons/component.svg",
+            "properties": ["checkboxProperty"],
+            "directiveOptions": {
+                "text": {
+                    "stripStylingOnPaste": false
```
wwjhu, line 24: I added the option here but noticed that there are not a lot of options set in the minimal sample. So is it ok to add it here or should we just rely on the individual validator tests?

lib/validate.ts
```diff
@@ -12,47 +12,50 @@ import { componentsDefinitionSchema_v1_2_x } from './components-schema-v1_2_x';
 import { componentsDefinitionSchema_v1_3_x } from './components-schema-v1_3_x';
 import { componentsDefinitionSchema_v1_4_x } from './components-schema-v1_4_x';
 import { componentsDefinitionSchema_v1_5_x } from './components-schema-v1_5_x';
+import { componentsDefinitionSchema_v1_6_x } from './components-schema-v1_6_x';
 
 import { parseDefinition } from './parser';
 import {
```
wwsno, line 18: Did you trigger your import sorting code :)?

test/resources/minimal-sample-1_6_x/components-definition.json
wwsno: Should be ok I think?

lib/validators/strip-styling-on-paste-validator.ts
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Validates whether stripStylingOnPaste is set for an editable directive.
+     */
+    validateComponent(parsedComponent: ParsedComponent): void {
+        if (!parsedComponent.directiveOptions) {
+            return;
+        }
+
+        for (const [key, directiveOptions] of Object.entries(parsedComponent.directiveOptions)) {
+            // Only for directives with stripStylingOnPaste option
```
wwsno, line 24: You can move the contents of this if block into a new function to reduce nesting and function length.

lib/validators/strip-styling-on-paste-validator.ts
```diff
@@ -0,0 +1,44 @@
...
+            if (!directive) {
+                this.error(
+                    `Component "${parsedComponent.name}" has stripStylingOnPaste set for an unknown directive "${key}".`,
+                );
+                continue;
+            }
+
+            if (directive.type !== 'editable') {
+                this.error(
+                    `Component "${parsedComponent.name}" has stripStylingOnPaste for a directive that is not editable "${key}". The type for that directive is set to "${directive.type}".`,
```
wwsno, line 39: "that is not of editable type" maybe? Not entirely clear that `editable` is the directive type in this error message I think.

test/validate.test.ts
```diff
@@ -72,24 +76,29 @@ describe('getValidators', () => {
     });
     it('should return amount of validators for version >= 1.0.0 and < 1.1.0', () => {
         const validators = getValidators('1.0.0', <any>null, <any>null, <any>null, <any>null, <any>null);
-        expect(validators && validators.length).toEqual(23);
+        expect(validators && validators.length).toEqual(24);
```
wwsno, line 79: 😅

lib/validators/directive-options-validator.ts
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Validates whether directiveOptions entries have a matching directive
+     */
+    validateComponent(parsedComponent: ParsedComponent): void {
+        if (!parsedComponent.directiveOptions) {
+            return;
+        }
+
+        for (const key of Object.keys(parsedComponent.directiveOptions)) {
+            if (!parsedComponent.directives[key]) {
```
wwsno, line 24: `for` block can be moved into a separate function to reduce nesting.

lib/validate.ts
wwsno: I would say yes, but could mean a confused customer if they have misconfigured directive options hanging around.

lib/components-schema-v1_6_x.ts
```diff
@@ -0,0 +1,600 @@
...
+                                        metadataField: {
+                                            type: 'string',
+                                            description:
+                                                'Metadata property name. The field is case sensitive. Slashes should be used to separate levels',
+                                        },
+                                    },
+                                },
+                                stripStylingOnPaste: {
+                                    type: 'boolean',
+                                    description: 'Strip all styling when pasting into the directive.',
```
wwsno, line 490: 'Strip all styling when pasting into an "editable" directive'? I believe we also don't end the descriptions with periods in the definition file in case of a single sentence, but seems inconsistent :(

wwsno (approved):

lib/validate.ts
wwjhu: I used "Optimize imports" because IntelliJ indicated some imports could be taken from barrel files and that sort too :-).

wwsno (approved):
🚀

PR #81, opened 2021-03-23
ECS-390: add slider control type property.

wwsno:
Requires configuring default value, minValue, maxValue and stepSize.

### DIFF
--- a/lib/validators/default-values-validator.ts
+++ b/lib/validators/default-values-validator.ts
@@ -8,20 +8,24 @@ import {
     ComponentPropertyControlCheckbox,
     ComponentPropertyControlRadio,
     ComponentPropertyControlSelect,
+    ComponentPropertyControlSlider,
     COMPONENT_PROPERTY_CONTROL_FITTING_VALUES,
 } from '../models/component-property-controls';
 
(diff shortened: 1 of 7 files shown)


lib/validators/default-values-validator.ts
```diff
@@ -90,10 +90,16 @@ export class DefaultValuesValidator extends Validator {
         return true;
     }
 
+    private validateNumberValue(property: ComponentProperty): boolean {
```
wwjhu, line 93: Should this be named `validateDefaultValue`?

wwjhu (approved):
Two minor thingies.

lib/validators/default-values-validator.ts
wwsno: Hmm, I think it is already clear this is about default values? This validates the default value is a number. The other two functions validate if they are a string or object. It depends on the control type which values are supported.

wwjhu: The way I read the function name is that it validates that a value is a number while in reality it validates that the default value is a number. But that same goes for the other functions that I didn't see, so let's leave it like this.

PR #88, opened 2021-03-26
Feature/component info restrict children test

ruudjungbacker:
For now this branch is based on the `feature/component-info` branch but I wanted to separate. A new test component set has been added that allows to test the restrict children path for the `generateComponentSetInfo` function.


test/resources/restrict-children-sample/localization/csCZ.json
```diff
@@ -0,0 +1,70 @@
+{
```
wwsno, line 1: You can leave out the localisation files for other languages than english in the test resources.

ruudjungbacker: Just removed them.

test/definition/component-info.test.ts
```diff
@@ -4,65 +4,148 @@ import { generateComponentSetInfo, processInfo } from '../../lib/definition/comp
...
+                                    type: 'editable',
+                                },
+                            ],
+                        },
+                    },
+                });
+            });
+        });
+
+        describe('processInfo', () => {
```
wwsno, line 61: Another processInfo describe block? That does not make sense. You should remove one.

wwsno (approved):
Looks ok.

jaapvanblaaderen (approved):
Looks good to me. Maybe also extend the sample in the other PR with a restrictChildren property (container for example)?

test/definition/component-info.test.ts
ruudjungbacker: Good point. Missed that one during a merge. Fixed it.

ruudjungbacker:
> Maybe also extend the sample in the other PR with a restrictChildren property (container for example)?

As far as I know this PR does so? Or am I missing an area?

PR #92, opened 2021-04-21
CSH-8048: only allow nameless properties for header properties.

wwsno:
The only type of property that should be allowed without name is the header, which does not save data
into the article. Other properties require a name, as it is used as the key for data saving.

I start to feel we should also have a fresh look at the validators structure:
- The tests feel a bit clumsy. I would rather have tools to create real component set structures in an easy way, rather than these incomplete snippets.
- The validators itself could use more structure, like I now need to pass through the ParsedComponent reference to build the error message.
- We need a way to easily show where the user made the mistake, with line numbers in the components-definition.json. Acting on the parsed ComponentSet structure might make this hard though.

### DIFF
--- a/lib/validators/properties-validator.ts
+++ b/lib/validators/properties-validator.ts
@@ -22,38 +22,29 @@ export class PropertiesValidator extends Validator {
         Object.values(this.componentSet.components).forEach((component) => this.validateComponent(component));
     }
 
-    /**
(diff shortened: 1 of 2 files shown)


wwjhu (approved):
👍

jaapvanblaaderen (approved):
👍

PR #93, opened 2021-05-04
ECS-395 parse child properties

wwjhu:
* Parses the child properties and builds the default component content when child properties are set.
* Adds JSON schema typing

Does not validate that child property names are unique for the component yet.

### DIFF
--- a/lib/components-schema-v1_7_x.ts
+++ b/lib/components-schema-v1_7_x.ts
@@ -6,7 +6,9 @@
 
 /* tslint:disable:max-line-length variable-name */
 
-const labelProperty = (description: string) => {
+import { JSONSchema7, JSONSchema7Definition } from 'json-schema';
+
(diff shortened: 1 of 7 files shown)


lib/components-schema-v1_7_x.ts
```diff
@@ -396,8 +400,51 @@ const componentPropertyDefinition = {
...
+    items: {
+        type: 'object',
+        properties: {
+            matchType: {
+                type: 'string',
+                description: `Defines how to match the parent property's value`,
+                enum: ['exact-value'],
+            },
+            matchExpression: {
+                type: ['boolean', 'integer', 'string', 'number'],
```
wwjhu, line 418: This is why `allowUnionTypes: true` is added.

wwjhu, line 418: Not entirely sure this is the right set of possible types.

wwsno (approved):
Lgtm 👍  🚀  🌔

PR #95, opened 2021-05-04
Restrict inline properties

wwjhu:
Restricts the inline properties in the component's properties array to just inline header property definitions and references to existing component properties overriding the directive key.
```
{ control: { type: 'header' }, label: 'Header Label' },
```
```
{ "name": "hyperlink", "directiveKey": "hyperlink" }
```

Adds validation that verifies that parsed header properties have no dataType property

### DIFF
--- a/lib/components-schema-v1_7_x.ts
+++ b/lib/components-schema-v1_7_x.ts
@@ -8,6 +8,11 @@
 
 import { JSONSchema7, JSONSchema7Definition } from 'json-schema';
 
+const defaultValue: JSONSchema7Definition = {
+    type: ['integer', 'string', 'number', 'object'],
+    description: 'Default value of property upon component creation. By default the property value is not defined.',
+};
+
 function labelProperty(description: string): { oneOf: JSONSchema7Definition[] } {
     return {
         oneOf: [
@@ -43,7 +48,7 @@ const inlineComponentPropertyDefinitionOrReferenceList: { oneOf: JSONSchema7Defi
         {
             type: 'object',
             additionalProperties: false,
-            required: ['name', 'directiveKey'],
+            anyOf: [{ required: ['name', 'directiveKey'] }, { required: ['name', 'defaultValue'] }],
             properties: {
                 name: {
                     type: 'string',
@@ -54,6 +59,7 @@ const inlineComponentPropertyDefinitionOrReferenceList: { oneOf: JSONSchema7Defi
                     type: 'string',
                     description: 'Directive key for properties that use a directive data type',
                 },
+                defaultValue: defaultValue,
             },
         },
         {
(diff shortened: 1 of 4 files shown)


lib/components-schema-v1_7_x.ts
```diff
@@ -35,6 +35,48 @@ function labelProperty(description: string): { oneOf: JSONSchema7Definition[] }
...
+            type: 'string',
+        },
+        {
+            type: 'object',
+            additionalProperties: false,
+            required: ['name', 'directiveKey'],
+            properties: {
+                name: {
+                    type: 'string',
+                    description: 'component property identifier',
```
jaapvanblaaderen, line 50: Small note: this description starts without capital, the other start with one. In the rest of the most descriptions start with a capital, though not all of them.

lib/components-schema-v1_7_x.ts
```diff
@@ -399,45 +441,30 @@ const componentPropertyDefinition: {
...
-                        {
-                            type: 'object',
-                            additionalProperties: false,
-                            // Use a copy to avoid a circular reference and a stack overflow on childProperties at run-time
-                            properties: { ...componentPropertyDefinition },
-                        },
-                    ],
+                matchType: {
+                    type: 'string',
+                    description: `Defines how to match the parent property's value`,
```
jaapvanblaaderen, line 452: Property's reads a bit weird. Maybe: "Defines how to match the value of the parent property"?

lib/components-schema-v1_7_x.ts
```diff
@@ -399,45 +441,30 @@ const componentPropertyDefinition: {
...
-            matchType: {
-                type: 'string',
-                description: `Defines how to match the parent property's value`,
-                enum: ['exact-value'],
-            },
-            matchExpression: {
-                type: ['boolean', 'integer', 'string', 'number'],
-                description: `The expression to use to match the parent property's value`,
-            },
+    childProperties: {
```
jaapvanblaaderen, line 444: Not really sure how this is done in the rest of the validator, but are/should there be tests to validate the schema? For example testing with a definition that contains an invalid `matchType`.

jaapvanblaaderen (approved):
LGTM, just a few remarks / questions

lib/components-schema-v1_7_x.ts
wwsno: I think we have tests in https://github.com/WoodWing/studio-component-set-tools/pull/93.

wwjhu: Positive flow test is indeed in #93, this PR restricts the possibilities. The question is what negative tests would bring us. Wouldn't that be just testing whether `AJV` does its job?

jaapvanblaaderen: Good point.

lib/components-schema-v1_7_x.ts
```diff
@@ -35,6 +35,48 @@ function labelProperty(description: string): { oneOf: JSONSchema7Definition[] }
...
 
+const inlineComponentPropertyDefinitionOrReferenceList: { oneOf: JSONSchema7Definition[] } = {
+    oneOf: [
+        {
+            type: 'string',
+        },
+        {
+            type: 'object',
+            additionalProperties: false,
+            required: ['name', 'directiveKey'],
```
wwsno, line 46: https://github.com/WoodWing/csde-components-boilerplate/blob/master/components/components-definition.json#L34

So we can also allow:
`{ "name": "prop", "defaultValue": "value" }`

wwjhu: Added.

wwsno (approved):

PR #96, opened 2021-05-05
ECS-395 validate child properties

wwjhu:
- restructures the tests (1st commit)
- adds child properties validation (2nd commit)

### DIFF
--- a/lib/validators/properties-validator.ts
+++ b/lib/validators/properties-validator.ts
@@ -26,6 +26,9 @@ export class PropertiesValidator extends Validator {
         const componentPropertyNames = new Set<string>();
 
         component.properties.forEach((property) => this.validateProperty(property, componentPropertyNames, component));
+        component.properties.forEach((property) =>
+            this.validateChildProperties(property, componentPropertyNames, component),
+        );
     }
 
     private validateProperty(
@@ -49,7 +52,7 @@ export class PropertiesValidator extends Validator {
         }
 
         if (componentPropertyNames.has(property.name)) {
-            this.error(`Component property "${property.name}" is not unique`);
+            this.error(`Component property "${property.name}" used in component "${component.name}" is not unique`);
         }
 
         componentPropertyNames.add(property.name);
@@ -87,5 +90,18 @@ export class PropertiesValidator extends Validator {
                 this.error(`Component properties "${property.name}" icon missing "${controlOption.icon}"`);
             }
         }
+    }
+
+    private validateChildProperties(
+        property: ComponentProperty,
+        componentPropertyNames: Set<string>,
+        component: ParsedComponent,
+    ) {
+        (property.childProperties || []).forEach((conditionalChildProperties) => {
+            const conditionalUsedProperyNames = new Set(componentPropertyNames);
+            (conditionalChildProperties.properties as ComponentProperty[]).forEach((childProperty) => {
+                this.validateProperty(childProperty, conditionalUsedProperyNames, component);
+            });
+        });
     }
 }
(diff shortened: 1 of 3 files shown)


lib/validators/properties-validator.ts
```diff
@@ -88,4 +91,17 @@ export class PropertiesValidator extends Validator {
...
         }
     }
+
+    private validateChildProperties(
+        property: ComponentProperty,
+        componentPropertyNames: Set<string>,
+        component: ParsedComponent,
+    ) {
+        (property.childProperties || []).forEach((conditionalChildProperties) => {
+            const conditionalUsedProperyNames = new Set(componentPropertyNames);
```
jaapvanblaaderen, line 101: Typo: `conditionalUsedPropertyNames`. This const can also be defined above the `forEach` I think?

test/validators/properties-validator.test.ts
```diff
@@ -1,121 +1,263 @@
...
-            expect(error).not.toHaveBeenCalled();
+            expect(errorSpy).not.toHaveBeenCalled();
         });
 
         it('should pass if there are multiple nameless properties of control type "header" in same component', () => {
-            definition.version = '1.5.0';
-            definition.components.c1.properties.push({ label: 'Header', control: { type: 'header' } });
-            definition.components.c1.properties.push({ label: 'Another Header', control: { type: 'header' } });
+            const { validator, errorSpy } = createPropertiesValidator({
+                version: '1.4.0',
```
jaapvanblaaderen, line 127: Shouldn't this be 1.5.0?

lib/validators/properties-validator.ts
```diff
@@ -88,4 +91,17 @@ export class PropertiesValidator extends Validator {
...
+            (conditionalChildProperties.properties as ComponentProperty[]).forEach((childProperty) => {
```
jaapvanblaaderen, line 102: Just wondering if this cast is safe when the property can also be a string? Didn't see any references to string based properties in the tests either?

test/validators/properties-validator.test.ts
```diff
@@ -1,121 +1,263 @@
...
+        label: 'header-label',
+    } as any;
+
+    function createConditionalProperty(children: any[]) {
+        return {
+            name: 'conditionalProperty',
+            control: {
+                type: 'text',
+            },
+            childProperties: [
```
wwsno, line 39: This is not the right construction, right? I would expect it has the matchExpression structure?

wwjhu: For these tests the matchExpression is not needed, but it now causes confusion

lib/validators/properties-validator.ts
wwjhu: It can't be moved, it needs to be reset for every conditional child properties object. Otherwise you can't use the same property in another conditional child properties object

lib/validators/properties-validator.ts
wwjhu: It is safe, the typing is incorrect. In `ParsedComponent`s these `properties` arrays have all been completely expanded with the definition from `componentProperties`

wwjhu: In `ParsedComponent`, `properties` is defined as `properties: ComponentProperty[];` whereas in `Component` it is defined as `properties?: (string | ComponentProperty)[];`

`childProperties` is part ComponentProperty as `properties: (string | ComponentProperty)[]`. Once parsed it is really down to `properties: ComponentProperty[]`. That would require a `ParsedComponentProperty` type similar to ParsedComponent. I quickly tried to make that work but kept running into compile errors and decided to revert

test/validators/properties-validator.test.ts
wwjhu: Added it

wwsno (approved):

PR #97, opened 2021-05-05
ECS-395: allow child properties in a defined subset of properties

wwjhu:


lib/validators/properties-validator.ts
```diff
@@ -13,6 +13,18 @@ import * as semver from 'semver';
 
 const RESERVED = [/^parallax$/];
 
+const TYPES_ALLOWING_CHILD_PROPERTIES = [
```
wwsno, line 16: Wonder if you could match the type to `ComponentPropertyControl['type'][]`? Would prevent typos.

https://github.com/WoodWing/studio-component-set-tools/blob/master/lib/models/component-property-controls.ts#L3

wwsno (approved):

jaapvanblaaderen (approved):
👌

lib/validators/properties-validator.ts
wwjhu: Good point

ruudjungbacker (approved):

PR #98, opened 2021-05-05
Release 1.13

wwjhu:
- Prepare for 1.8 schema
- Bump version to 1.13


wwsno (changes requested):
We are not releasing 1.7.0 yet. We are first releasing it under "1.7.0-next" so we can test out the conditional properties feature.

wwjhu:
Ok, so roll back the 1.8.0-next changes. When we're comfortable with the changes we would release 1.13.1, or 1.14?

wwsno:
I guess we could release `1.13.1`. Not sure what the best approach is there, doesn't matter too much.

wwsno (approved):

PR #101, opened 2021-05-10
ECS-392: Disallow child properties to have a default value

wwjhu:
Only on the top level allow default values to prevent problems in the editor. Alternatively we can support to editor for having (different) default values in child properties, but let's wait till that is requested.

### DIFF
--- a/lib/components-schema-v1_7_x.ts
+++ b/lib/components-schema-v1_7_x.ts
@@ -40,48 +40,54 @@ function labelProperty(description: string): { oneOf: JSONSchema7Definition[] } 
     };
 }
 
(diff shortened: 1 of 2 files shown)


wwsno (approved):
Lgtm 💯

wwsno:
> I guess this change conflicts with the default values I specified for [WoodWing/apple-news-component-sets#39](https://github.com/WoodWing/apple-news-component-sets/pull/39)? Especially for the sliders, you need a default value.

It's still possible, but only allowed in the root. The editor currently does not support default values in child properties and we don't need it right now.

PR #102, opened 2021-05-11
ECS-408: Allow boolean values for checkbox property

jaapvanblaaderen:
Boolean values can be safely used when dataType equals "data" as they are not used in css. This change allows cleaner definition of the checkboxes we need in the Apple News component set.

### DIFF
--- a/lib/validators/properties-validator.ts
+++ b/lib/validators/properties-validator.ts
@@ -111,7 +111,9 @@ export class PropertiesValidator extends Validator {
             return;
         }
         if (typeof property.control.value === 'boolean' && property.dataType !== 'data') {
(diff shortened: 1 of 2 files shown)


lib/validators/properties-validator.ts
```diff
@@ -105,6 +106,17 @@ export class PropertiesValidator extends Validator {
         }
     }
 
+    private validateCheckBoxValue(property: ComponentProperty) {
```
jaapvanblaaderen, line 109: I wasn't really sure if this validation needs to be here or inside the schema. I guess we might be able to validate this with JSONSchema, but that might make the schema a bit complex/ugly.

wwsno (approved):

PR #105, opened 2021-05-20
ECS-409: Add custom style support in the component set package

ruudjungbacker:


lib/components-schema-v1_8_x.ts
```diff
@@ -665,6 +665,27 @@ export const componentsDefinitionSchema_v1_8_x: JSONSchema7 = {
...
+                'List of custom styles definitions which can be configured in the look and feel style of an article',
+            items: {
+                type: 'object',
+                properties: {
+                    label: { type: 'string', minLength: 1, maxLength: 20 },
+                    key: {
+                        type: 'string',
+                        minLength: 1,
+                        maxLength: 20,
+                        pattern: '^[a-zA-Z0-9-_]*$',
```
ruudjungbacker, line 681: I'm not sure if these types of checks should be part of the schema or the validator?

lib/validate.ts
```diff
@@ -311,6 +312,7 @@ export function getValidators(
             new UnitTypeValidator(error, componentSet),
             new LocalizationValidator(error, componentSet, filePaths, getFileContent),
             new PackageValidator(error, componentSet, filePaths, getFileSize),
+            new CustomStylesValidator(error, componentSet, filePaths),
```
ruudjungbacker, line 315: For now I added the validator to all versions. The advantage of this is that the validator checks all the uploads for files in the `styles/customStyles` directory.

lib/components-schema-v1_8_x.ts
```diff
@@ -665,6 +665,27 @@ export const componentsDefinitionSchema_v1_8_x: JSONSchema7 = {
...
         },
+
+        customStyles: {
+            type: 'array',
+            description:
+                'List of custom styles definitions which can be configured in the look and feel style of an article',
+            items: {
+                type: 'object',
+                properties: {
+                    label: { type: 'string', minLength: 1, maxLength: 20 },
```
ruudjungbacker, line 676: The label is now defined in the component set definition. Should it be possible to translate this property?

lib/components-schema-v1_8_x.ts
```diff
@@ -665,6 +665,27 @@ export const componentsDefinitionSchema_v1_8_x: JSONSchema7 = {
                 type: 'string',
             },
         },
+
```
ruudjungbacker, line 668: Added this part to the 1.8 next schema. Is this fine or should a new 1.9 schema already be added?

wwsno: Yes, that's the purpose of the "-next" schema: https://github.com/WoodWing/studio-component-set-tools#releasing-a-new-schema-version

In case the readme is not clear enough, maybe improve it?

lib/components-schema-v1_8_x.ts
wwsno: I'm ok with it not being translatable.

test/parser/parser-util.test.ts
```diff
@@ -298,6 +298,7 @@ describe('Parser utils', () => {
                         },
                     },
                     scripts: [],
+                    customStyles: [],
```
wwsno, line 301: It's optional right?

lib/validators/custom-styles-validator.ts
```diff
@@ -0,0 +1,33 @@
...
+            }
+
+            if (!this.filePaths.has(path.normalize(customStyle.default))) {
+                this.error(`The default file for custom style "${customStyle.label}" does not exist`);
+            }
+        });
+    }
+
+    private checkCustomStylesFolder() {
+        if (Array.from(this.filePaths).some((file) => file.toLowerCase().startsWith('styles/customstyles/'))) {
```
wwsno, line 29: So the customer can't provide a default config?

test/validators/custom-styles-validator.test.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    describe('validate', () => {
+        it('should do nothing when no custom styles are given', async () => {
+            delete definition.customStyles;
+            validator = new CustomStylesValidator(error, definition, new Set());
+            await validator.validate();
+            expect(error).not.toHaveBeenCalled();
+        });
+
+        it('should do nothing when no custom styles are given', async () => {
```
wwsno, line 25: Same test description as previous one, but this one is for an empty array.

test/validators/custom-styles-validator.test.ts
```diff
@@ -0,0 +1,83 @@
...
+                key: 'test-style',
+                label: 'Test Style',
+                type: 'JSON',
+            });
+            validator = new CustomStylesValidator(error, definition, new Set());
+            await validator.validate();
+            expect(error).not.toHaveBeenCalled();
+        });
+
+        it(`should throw now errors when the default file is available`, async () => {
```
wwsno, line 42: `now errors` -> `no errors`.

test/validators/custom-styles-validator.test.ts
```diff
@@ -0,0 +1,83 @@
...
+            definition.customStyles.push({
+                key: 'test-style',
+                label: 'Test Style',
+                type: 'JSON',
+                default: './styles/default-custom-styles/existing-file.json',
```
wwsno, line 47: Wouldn't it make more sense if this is placed in the `customStyles ` folder? Also naming is inconsistent. One has dashes, other one camelCase.

lib/components-schema-v1_8_x.ts
wwjhu: This is fine

lib/validators/custom-styles-validator.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    async validate(): Promise<void> {
+        this.checkCustomStylesFolder();
+
+        if (!this.componentSet.customStyles) {
+            return;
+        }
+
+        this.componentSet.customStyles.forEach((customStyle) => {
```
wwjhu, line 17: This part should be in a separate `validate` function so that it explains by its name what it is doing.

wwjhu (approved):

lib/components-schema-v1_8_x.ts
ruudjungbacker: I think the readme is clear. But I didn't know in what version the conditional properties were released. As those are in 1.7.0 it turns out this feature is actually 1.8.0.

test/parser/parser-util.test.ts
ruudjungbacker: Yes it is. I used the same structure as the scripts in the parser-utils. https://github.com/WoodWing/studio-component-set-tools/pull/105/files#diff-b3b353126cc79046a828b28e511fd0a613b94cfd96a7ff84a18ea8f69b8f7ab4R48 So the parsed component definition always contains the property with an empty array. Is this what we want?

wwsno: This is fine.

lib/validators/custom-styles-validator.ts
ruudjungbacker: They can. The default property should be a file path to any file in the component set. For example, with Apple News the default property will point to the `custom/apple-news/styles.json` file. This means we can already upgrade the component set. Once the UI is done you should be able to update the styles then. 

The backend however stores the custom styles within the styles folder. So when the component set already contains a styles/customStyles folder it will be very hard to merge and clean the directory. For now I decided that this folder just can't be used in the component set. The default styles can be stored anywhere else. Including in a customStyles folder in the root of the component set.

lib/validators/custom-styles-validator.ts
ruudjungbacker: I just added the new function.

test/validators/custom-styles-validator.test.ts
ruudjungbacker: Ok, only the empty array one should be tested. The `parseDefinition` makes sure it is always an array. So this should be fine.

test/validators/custom-styles-validator.test.ts
ruudjungbacker: The default styles aren't allowed in the `styles/customStyles` folder for syncing reasons. I now added them to a defaultCustomStyles folder in the root which should be fine. The reason for this logic was this comment: https://github.com/WoodWing/ecs-backend/pull/874#pullrequestreview-661823134

wwsno (approved):
👍

PR #111, opened 2021-08-10
CSH-8342: Add dynamicCaptions property

wwjhu:
If I am not mistaken, there is nothing more to it than this.


jaapvanblaaderen (approved):
I guess you're right, there is no real behavior that needs additional testing next to schema testing I assume?

kylejwatson (approved):
👍

wwsno (approved):
👍

PR #114, opened 2021-08-24
CSH-8382: Add windows runner

wwsno:
Run tests on Windows. One failing test due normalisation of paths. I do wonder how we want to deal with path normalisation. In the backend we always want to store the paths the same way.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,14 @@ on: push
 
 jobs:
     build:
-        runs-on: ubuntu-latest
+        runs-on: ${{ matrix.os }}
+        strategy:
+            matrix:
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -3,7 +3,14 @@ on: push
...
 jobs:
     build:
-        runs-on: ubuntu-latest
+        runs-on: ${{ matrix.os }}
+        strategy:
+            matrix:
+                os: [ubuntu-latest, windows-latest]
+        defaults:
+            run:
+                shell: ${{ matrix.os == 'windows-latest' && 'cmd' || 'bash' }}
```
wwsno, line 12: On windows the default shell is apparently powershell and it does not make the build fail on non zero exit codes...

.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
wwsno, line 1: prettier/prettier#7825

wwjhu (approved):
👍

PR #115, opened 2021-08-24
CSH-8382: remove path.normalise in parser util.

wwsno:
The getFileContent function can take care of normalising the path if it wants.
By not calling path here, we should not need to include a path polyfill in our code.


kylejwatson (approved):
👍

jaapvanblaaderen (approved):

wwjhu (approved):

PR #119, opened 2021-09-07
CSH-8344: type guards

wwjhu:
- Add typeguards for all `ComponentPropertyControl` types
- Add behavior typeguard for component property controls that support dynamic captions (currently just select)


wwsno (approved):
Ok.

PR #123, opened 2021-09-07
scout: make typing more flexible

wwjhu:
Makes it possible to filter for properties with a specific type of control and add a matching type.


wwsno (approved):
LGMT 👍  💯

PR #125, opened 2021-09-16
scout: Unmark properties as optional

wwjhu:
These are optional in `ComponentsDefinition`, but `parseDefinition` guarantees that they always exist in `ComponentSet`.


wwsno (approved):
👍

PR #131, opened 2021-09-21
scout: tighten ts code checking

wwjhu:
Noticed warnings about unused imports in the IDE. Turns them into errors through TS.


ruudjungbacker (approved):
🥳

PR #132, opened 2021-09-21
CSH-8382: roll back 523b8ba and solve in a different way

wwjhu:
Fixes the problem in ecs backend where the filePaths and getFileContent paths no longer match because one is normalized and the other isn't. Also keeps the getFileContent, getFileSize and filePaths arguments of the validate function consistent: all take normalized paths. The problem of having a path.normalize in the browser is solved by separating the loading of the renditions is a separate function that is not imported in the browser when using parseDefinition. Additionally adds deepFreeze to ensure the components definition is not altered while parsing it into a component set.

### DIFF
--- a/lib/models/get-file-content.ts
+++ b/lib/models/get-file-content.ts
@@ -6,9 +6,12 @@ export type GetFileContentOptionsType = { encoding: string | null } | null;
 /**
  * Return string if encoding is string and Buffer otherwise.
  */
-export type GetFileContentType = (filePath: string, options?: GetFileContentOptionsType) => Promise<string | Buffer>;
+export type GetFileContentType = (
+    normalizedRelativeFilePath: string,
+    options?: GetFileContentOptionsType,
+) => Promise<string | Buffer>;
 
 /**
  * Return file size in bytes.
  */
-export type GetFileSize = (filePath: string) => Promise<number>;
+export type GetFileSize = (normalizedRelativeFilePath: string) => Promise<number>;
--- a/lib/renditions/utils.ts
+++ b/lib/renditions/utils.ts
@@ -1,13 +1,13 @@
 import * as path from 'path';
 import { Component, ComponentRendition, ComponentsDefinition, GetFileContentType } from '../models';
-import merge = require('lodash.merge');
+import cloneDeep = require('lodash.clonedeep');
 import { deepFreeze } from '../util/freeze';
 
 export async function loadHtmlRenditions(
     componentsDefinition: ComponentsDefinition,
     getFileContent: GetFileContentType,
 ): Promise<ComponentsDefinition> {
-    const enrichedDefinition = merge({}, componentsDefinition);
+    const enrichedDefinition = cloneDeep(componentsDefinition);
     for (const compDef of enrichedDefinition.components) {
         if (!hasRendition(compDef, ComponentRendition.HTML)) {
             await loadRendition(compDef, ComponentRendition.HTML, getFileContent);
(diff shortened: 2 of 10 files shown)


lib/models/get-file-content.ts
```diff
@@ -6,9 +6,12 @@ export type GetFileContentOptionsType = { encoding: string | null } | null;
 /**
  * Return string if encoding is string and Buffer otherwise.
  */
-export type GetFileContentType = (filePath: string, options?: GetFileContentOptionsType) => Promise<string | Buffer>;
+export type GetFileContentType = (
+    normalizedRelativeFilePath: string,
```
ruudjungbacker, line 10: I like the changes of these parameter names. But should they also be updated in the readme? The readme only mentions:
`const getFileContent = async (filePath: string) => 'fileContent';`

It doesn't mention the options at all (which isn't that important I think). but the name could be updated to normalizedRelativeFilePath maybe?

lib/renditions/utils.ts
```diff
@@ -0,0 +1,38 @@
+import * as path from 'path';
+import { Component, ComponentRendition, ComponentsDefinition, GetFileContentType } from '../models';
+import merge = require('lodash.merge');
+import { deepFreeze } from '../util/freeze';
+
+export async function loadRenditions(
```
ruudjungbacker, line 6: Not sure, but shouldn't this be `loadHtmlRenditions`? as the PSV one is left out of scope.

lib/renditions/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+async function loadRendition(
+    component: Component,
+    rendition: ComponentRendition,
+    getFileContent: GetFileContentType,
+): Promise<Component> {
+    if (!component.renditions) {
+        component.renditions = {};
+    }
+    component.renditions[rendition] = (await getFileContent(
+        path.normalize(`./templates/${rendition}/${component.name}.html`),
```
ruudjungbacker, line 32:
```suggestion
        path.normalize(`templates/${rendition}/${component.name}.html`),
```

lib/renditions/utils.ts
```diff
@@ -0,0 +1,38 @@
+import * as path from 'path';
+import { Component, ComponentRendition, ComponentsDefinition, GetFileContentType } from '../models';
+import merge = require('lodash.merge');
+import { deepFreeze } from '../util/freeze';
+
+export async function loadRenditions(
+    componentsDefinition: ComponentsDefinition,
+    getFileContent: GetFileContentType,
+): Promise<ComponentsDefinition> {
+    const enrichedDefinition = merge({}, componentsDefinition);
```
ruudjungbacker, line 10: Just for my understanding, this 'unfreezes' an object?

ruudjungbacker (commented):
I added a few comments/questions, but I like the overal solution for the issue!

lib/renditions/utils.ts
wwjhu: Makes sense.

lib/renditions/utils.ts
wwjhu: It creates a [copy of the object](https://lodash.com/docs#merge)

ruudjungbacker (approved):
🚀

lib/renditions/utils.ts
wwsno: Shouldn't you use lodash deepClone in this case? Or is it different?

wwsno (approved):
Looks fine 👍

lib/renditions/utils.ts
wwjhu: The result is the same. I copied this from the old code, but I guess that for new code we're better off using the function that describes best what it does: cloneDeep.

PR #135, opened 2021-09-28
chore: update jest to version 27.

wwsno:
Jest now uses jest-circus by default instead of jasmine. This required replacing some incorrect code in the tests.


wwjhu (approved):
👍

PR #137, opened 2021-10-08
Csh 6552 improve types

wwsno:
- Rename `Component` to `ComponentDefinition` 
- Rename `ParsedComponent` to `Component` (which is really the main type we would use).

I checked usage of `ParsedComponent` in the Digital Editor and not too many direct usages. So updating should be trivial.

As per ticket I also looked if we could change `Label` to `string` for the final type of the component set, as in the client it is always already translated. However this seems to result in a mess of types (not trivial to keep clean).

### DIFF
--- a/lib/validators/doc-container-groups-validator.ts
+++ b/lib/validators/doc-container-groups-validator.ts
@@ -7,7 +7,7 @@
  */
 
 import { Validator } from './validator';
-import { ParsedComponent } from '../models';
+import { Component } from '../models';
 import { GroupsValidator } from './groups-validator';
 
 export class DocContainerGroupsValidator extends Validator {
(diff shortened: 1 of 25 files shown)


lib/validators/doc-container-groups-validator.ts
```diff
@@ -26,7 +26,7 @@ export class DocContainerGroupsValidator extends Validator {
      * @param errorReporter
      * @param parsedComponent
      */
-    validateComponent(parsedComponent: ParsedComponent): void {
+    validateComponent(parsedComponent: Component): void {
```
jaapvanblaaderen, line 29: There are multiple instances where the variable is still called `parsedComponent`. I guess we want to rename those as well?

wwsno: Done 👍

jaapvanblaaderen (approved):
Noice! 🚀 🧇 🍪 💯 🎉

PR #138, opened 2021-10-08
Remove lodash usage

wwsno:
This PR removes the usage of lodash from the library:
- The used packages were last updated over 3 years ago.
- The main lodash package tends to cause security vulnerability warnings (annoying).

Instead rewrite the code to work without merge and cloneDeep. A few shallow clones are enough.

For `validate` entry point this should be  safe refactor, as it performs a deep freeze of the components definition object. This entry point also calls the parser utility. So the tests would quickly complain about not being able to assign properties.

In addition not doing a full clone everywhere is likely more performant as well (can be useful for our tests that need to parse the component definition).

### DIFF
--- a/lib/parser/parser-utils.ts
+++ b/lib/parser/parser-utils.ts
@@ -38,11 +38,10 @@ export async function parseDefinition(componentsDefinition: ComponentsDefinition
         customStyles: componentsDefinition.customStyles || [],
     };
     // copy source because properties and child properties will be expanded
-    const definition: ComponentsDefinition = JSON.parse(JSON.stringify(componentsDefinition));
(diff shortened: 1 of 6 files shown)


jaapvanblaaderen (approved):
LGTM

lib/parser/parser-utils.ts
```diff
@@ -39,11 +38,10 @@ export async function parseDefinition(componentsDefinition: ComponentsDefinition
         customStyles: componentsDefinition.customStyles || [],
     };
     // copy source because properties and child properties will be expanded
```
ruudjungbacker, line 40: Is this comment still valid?

PR #143, opened 2021-10-14
Ecs 441 character styles

wwsno:
Draft definition for character styles.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@woodwing/studio-component-set-tools",
-  "version": "1.16.4",
+  "version": "1.16.5",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@woodwing/studio-component-set-tools",
-  "version": "1.16.4",
+  "version": "1.16.5",
   "description": "Validation module for Studio Digital Editor component sets.",
   "author": "",
   "license": "Apache-2.0",


test/validate.test.ts
```diff
@@ -1,7 +1,29 @@
...
+import { validate, validateFolder, getValidators, validatePackageSize, readFile, getSize } from '../lib/validate';
 import * as chalk from 'chalk';
+import { listFilesRelativeToFolder } from '../lib/util/files';
+import { GetFileContentOptionsType } from '../lib/models';
 
 describe('validateFolder', () => {
+    function createValidator(fileCustomiser: (filePath: string, content: string) => string) {
+        const errorSpy = jest.fn();
+
+        async function validateFolderWithCustomiser(folderPath: string): Promise<boolean> {
```
wwsno, line 11: Experimenting a bit with customising the components definition with alternative scenarios. Not sure yet if this is the way to go, but maybe it is a start.

lib/components-schema-v1_9_x.ts
```diff
@@ -690,6 +690,18 @@ export const componentsDefinitionSchema_v1_9_x: JSONSchema7 = {
...
         },
+
+        characterStyles: {
+            type: 'array',
+            description: 'List of character style definitions which can be applied to text.',
+            items: {
+                type: 'object',
+                properties: {
+                    label: { type: 'string' },
+                    id: { type: 'string', pattern: '^cs--?[_a-zA-Z]+[_a-zA-Z0-9-]*' },
```
wwjhu, line 701: With our `cs-` prefix in place the pattern can be simpler:
```
id: { type: 'string', pattern: '^cs-[_a-zA-Z0-9-]+' },
```

PR #145, opened 2021-12-06
CSH-8655: add experimental property definition for selecting objects (such as images).

wwsno:
For this task review if the initial API design is good enough.


kylejwatson (approved):
Think this makes sense with what was discussed, allows for flexibility, and 'object' sounds like a fair name

wwjhu (approved):
lgtm

PR #148, opened 2021-12-10
boyscout: restrict data type for new object-select property.

wwsno:
Doesn't make sense as "styles" or "inlineStyles", as in that case it will be added as styling to the dom. But the data is an object, so that would just result in weird behavior.


kylejwatson (approved):
Makes sense 👍

PR #149, opened 2021-12-13
CSH-8707: introduce object data type.

wwsno:
This is used to indicate the data refers to an object in Studio.
When a property is created with this data type, it ensures placed relations are created with that object.
For now it can only be used with the experimental "object-select" UI control type.
Archiving to assets is not supported (the image will not be picked up).

### DIFF
--- a/lib/models/components-definition.ts
+++ b/lib/models/components-definition.ts
@@ -148,6 +148,7 @@ export interface ComponentProperty<ControlType = ComponentPropertyControl> {
         | 'styles'
         | 'inlineStyles'
         | 'data'
+        | 'object'
         | 'doc-editable'
         | 'doc-image'
         | 'doc-html'
--- a/lib/models/components-definition.ts
+++ b/lib/models/components-definition.ts
@@ -148,7 +148,7 @@ export interface ComponentProperty<ControlType = ComponentPropertyControl> {
         | 'styles'
         | 'inlineStyles'
         | 'data'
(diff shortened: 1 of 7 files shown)


lib/models/components-definition.ts
```diff
@@ -148,6 +148,7 @@ export interface ComponentProperty<ControlType = ComponentPropertyControl> {
         | 'styles'
         | 'inlineStyles'
         | 'data'
+        | 'object'
```
jaapvanblaaderen, line 151: This is obviously correct but also confusing (we should have never used the name 'object' in Studio/Enterprise). Something like `studio-object` might also be ugly though.

jaapvanblaaderen (approved):
LGTM, just don't like 'object'

lib/models/components-definition.ts
wwsno: You might be right that `studio-object` communicates it more clearly. What do others think?

kylejwatson: I think `studio-object` makes sense for this case. As long as it isn't reused if the object-select control was sourcing 'objects' from somewhere other than studio

wwjhu: Make sense to name it `studio-object`. The accepted values of the properties in the control are also quite linked to Studio at the moment.

wwjhu (approved):
👍

PR #151, opened 2021-12-23
chore: make 1.9.0 final

wwsno:
- Support character styles being defined
- Support studio-object & studio-object-select (experimental in UI).


wwjhu (approved):
lgtm

PR #155, opened 2022-01-04
boyscout: define required properties of character styles.

wwsno:
The validator would not complain when leaving out properties or when making typos.


kylejwatson (approved):
Nice 👍

PR #157, opened 2022-02-25
CSH-8877: Add component property info to the partial info definition

jaapvanblaaderen:
Information will be used in the Studio plug-in while parsing the article to find all `studio-object-select` properties.

### DIFF
--- a/lib/definition/component-info.ts
+++ b/lib/definition/component-info.ts
@@ -1,12 +1,18 @@
-import { ComponentsDefinition, ComponentDefinition, ComponentRendition } from '../models/components-definition';
+import {
+    ComponentsDefinition,
+    ComponentDefinition,
+    ComponentRendition,
+    ComponentProperty,
+} from '../models/components-definition';
 import {
     ComponentField,
     ComponentSetInfo,
     ComponentInfoFields,
-    ComponentInfoProperties,
+    ComponentInfoProperty,
 } from '../models/component-set-info';
 import { RenditionResolver, processTemplates } from './process-templates';
 import parse5 = require('parse5');
+import { parseDefinition } from '../parser';
 
 const directivePrefix = 'doc-';
 
@@ -30,20 +36,16 @@ export async function generateComponentSetInfo(
  * This implementation can be used when the components definition already contains the rendition information.
  */
 export function processInfo(componentsDefinition: ComponentsDefinition): ComponentSetInfo {
+    const componentSet = parseDefinition(componentsDefinition);
     return {
-        components: componentsDefinition.components.reduce((result, component) => {
+        components: Object.values(componentSet.components).reduce((result, component) => {
             result[component.name] = {
                 fields: parseFields((<{ html: string }>component.renditions)[ComponentRendition.HTML]),
+                properties: parseProperties(component.properties),
             };
             result[component.name].fields.forEach((f) => addRestrictChildrenInfo(componentsDefinition, component, f));
             return result;
         }, {} as ComponentInfoFields),
-        componentProperties: componentsDefinition.componentProperties.reduce((result, property) => {
-            result[property.name] = {
-                dataType: property.dataType,
-            };
-            return result;
-        }, {} as ComponentInfoProperties),
     };
 }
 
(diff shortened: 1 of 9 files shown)


lib/models/component-set-info.ts
```diff
@@ -1,43 +1,25 @@
...
- *      fields:    [
- *          {
- *              type: "editable",
- *              contentKey: "text"
- *          }
- *      ]
- *   }
- *   ... etc ...
- * }
- */
```
jaapvanblaaderen, line 24: Removed this as it was incorrect and also duplicated in the README

lib/definition/component-info.ts
```diff
@@ -33,6 +38,12 @@ export function processInfo(componentsDefinition: ComponentsDefinition): Compone
             result[component.name].fields.forEach((f) => addRestrictChildrenInfo(componentsDefinition, component, f));
             return result;
         }, {} as ComponentInfoFields),
+        componentProperties: componentsDefinition.componentProperties.reduce((result, property) => {
```
wwsno, line 41: I'm not sure this is good enough. The full definition parses properties per component:
https://github.com/WoodWing/studio-component-set-tools/blob/master/lib/parser/parser-utils.ts#L130

It might be possible to override the dataType in the component property.

jaapvanblaaderen: It was (kind of) working as overriding the dataType is (currently) not allowed. But you are obviously right that this is not the correct approach so I rewrote it by using the parsed component set and specifying the properties per component.

lib/definition/component-info.ts
```diff
@@ -93,3 +106,12 @@ function addRestrictChildrenInfo(
     }
     field.restrictChildren = allowedComponentsNames;
 }
+
+function parseProperties(properties: ComponentProperty[]) {
+    return properties.reduce((infoProperties, property) => {
+        if (property.name) {
```
jaapvanblaaderen, line 112: I added this condition to rule out header properties, stumbled upon this by accident as there was a header property in a slideshow component of one of the tests. Not really sure if this is the best way to check it though? I was also wondering if it's correct that the component property `name` is currently [not optional](https://github.com/WoodWing/studio-component-set-tools/blob/master/lib/models/components-definition.ts#L136)?

lib/validate.ts
```diff
@@ -174,9 +174,13 @@ export async function validate(
...
     try {
-        componentSet = await parseDefinition(await loadHtmlRenditions(componentsDefinition, getFileContent));
+        componentSet = parseDefinition(await loadHtmlRenditions(componentsDefinition, getFileContent));
     } catch (e) {
-        errorReporter(e.message ?? e);
+        if (!(e instanceof Error)) {
+            // Re-throw any non Error objects as these should not happen (software error, not a component set issue)
+            throw e;
+        }
+        errorReporter(e.message);
```
jaapvanblaaderen, line 183: Had to make these changes as `parseDefinition` does not contain any async logic and we don't want to make the `processInfo` function async.

lib/definition/component-info.ts
wwsno: `header` indeed ended up being weird. Maybe check on the `control.type`? According to the json schema it is not optional, but `header` bypasses it because it is merged in...
It's an ugly exception.

lib/validate.ts
wwsno: hmm, I'm a bit unsure. Eventually we might add async logic to the parsing?

test/definition/component-info.test.ts
```diff
@@ -43,6 +43,12 @@ describe('ComponentInfo', () => {
                                 type: 'editable',
                             },
                         ],
+                        properties: [
```
wwsno, line 46: I guess you could also opt for a structure like:
```typescript
{
    selectProperty: {
        dataType: 'styles',
    }
}
```

lib/validate.ts
jaapvanblaaderen: It just feels weird to make these functions async when they not contain async logic. But from an "API" perspective allowing async logic in the future I guess it could make sense.

test/definition/component-info.test.ts
jaapvanblaaderen: Yeah, was already doubting what would be better. But I guess this indeed better aligns with the way the components are listed here as well (and allows for quick lookups).

wwsno: It depends on our goal I think. The order of the properties is also the order it is appearing in the component properties panel.

jaapvanblaaderen: I guess that's only important when you want to build a properties UI? In that case you need the full definition anyway. My assumption is that the component info is used by article processing tools to correctly interpret the article, allowing conversion / data lookup.

wwsno: Indeed 👍

wwsno (approved):

PR #159, opened 2022-03-11
CSH-8877: Include child properties in the list of component info properties

jaapvanblaaderen:
Using the `minimal-component-set-next` as test component set as that already includes child properties.


wwsno (approved):
LGTM.

I wonder if we could share the code with the digital-editor-core, but not too important for now.

PR #164, opened 2022-08-10
Feature/csh 9204 anchor property type

wwsno:
- Introduce "anchor" control type (experimental)
- Support child properties and data type "data"
- Available in "1.10.0-next" (not implemented yet)

When the control is added, you will have two buttons to focus or delete the created anchor (creation happens through the text toolbar). Similar to Inception:
Screenshot 2022-08-10 at 08 34 38

The properties from Inception should be defined as child properties. For this a new match type is introduced "any-value", as we don't have a value to perform an exact match.

### DIFF
--- a/test/validators/properties-validator.test.ts
+++ b/test/validators/properties-validator.test.ts
@@ -63,7 +63,18 @@ describe('PropertiesValidator', () => {
         };
     }
 
-    function createConditionalProperty(children: any[]) {
+    function createAnchorProperty(): ComponentProperty {
+        return {
+            name: 'anchorProperty',
+            label: 'Anchor',
+            control: {
+                type: 'anchor',
+            },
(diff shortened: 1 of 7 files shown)


test/validators/properties-validator.test.ts
```diff
@@ -388,4 +399,47 @@ describe('PropertiesValidator', () => {
             expect(errorSpy).toHaveBeenCalledWith(`Cannot use dataType "studio-object" with control type "text"`);
         });
     });
+
+    describe('anchor property', () => {
+        it('should pass with dataType "studio-object"', () => {
```
wwjhu, line 404: studio-object?

wwjhu (commented):
Looks good!

jaapvanblaaderen (approved):
LGTM

PR #165, opened 2022-08-11
Bump validator version to 1.20.0

wwsno:
Include experimental definition for "anchor" property and adds "any-value" match type. Available in component definition schema 1.10.0-next.


kylejwatson (approved):
:rocket: :bookmark:

PR #166, opened 2022-08-23
boyscout: Split out component conversion rule types

jaapvanblaaderen:
Makes it easier to work with the various types, to be used in: https://github.com/WoodWing/contentstation/pull/3736


wwsno (approved):
👍

PR #176, opened 2023-08-18
Csh 9935 infogram directive

Gecete:
Added doc-infogram directive to be able to work with infogram tables in digital editor. This directive allows to set properties such as `logoPath` and `link` for the infogram logo at the bottom right of the components panel.


lib/validate.ts
```diff
@@ -289,8 +291,10 @@ function getValidationSchemaSource(version: string): Schema | null {
         return componentsDefinitionSchema_v1_9_x;
     } else if (semver.satisfies(version, '1.10.0', semVerOptions)) {
         return componentsDefinitionSchema_v1_10_x;
-    } else if (semver.satisfies(version, '1.11.0-next', semVerOptions)) {
+    } else if (semver.satisfies(version, '1.11.0', semVerOptions)) {
```
wwjhu, line 294: This is not how the versioning works for `studio-component-set-tools`. The tables feature will be part 1.11 (which is the version under development at the moment). See https://github.com/WoodWing/studio-component-set-tools/blob/f2be3159eeff8cb0d4f607edd0aac9fbacc9f744/README.md#releasing-a-new-schema-version.

Also, this kind of change needs to be in a separate PR.

lib/components-schema-v1_11_x.ts
```diff
@@ -472,6 +490,7 @@ const componentPropertyDefinition: {
             'doc-html',
             'doc-slideshow',
             'doc-media',
+            'doc-infogram',
```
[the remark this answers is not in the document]
Gecete: I originally called it `doc-table` but Sander suggested to use infogram as it supports more than tables... maybe `doc-infographic`?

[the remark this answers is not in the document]
wwjhu: Maybe `doc-chart`? They talk about chart types, so chart seems to be the overarching term?

image

lib/components-schema-v1_11_x.ts
```diff
@@ -101,6 +101,10 @@ const componentGroupDefinition: JSONSchema7Definition = {
                 minLength: 3,
             },
             label: labelProperty('Group label shown in Digital Editor'),
+            integrationLogo: {
```
wwjhu, line 104: Here we use it for an integration but this could be a group logo for other groups too, don't you agree?

lib/components-schema-v1_11_x.ts
```diff
@@ -101,6 +101,10 @@ const componentGroupDefinition: JSONSchema7Definition = {
...
+                type: 'object',
+                description: ' Display logo of integration',
```
wwjhu, line 106:
```suggestion
                description: 'Group logo with link',
```

lib/models/component-property-controls.ts
```diff
@@ -213,6 +214,18 @@ export function isMediaProperties(
     return control.type === 'media-properties';
 }
 
+/**
+ * Chart properties control.
+ * Contains parameters to set the chart logo
```
wwjhu, line 219: Comment is outdated

lib/models/components-definition.ts
```diff
@@ -186,12 +187,20 @@ export interface ComponentGroup {
...
+    /** Display logo of integration */
+    integrationLogo?: IntegrationLogo;
 }
 
 export interface ComponentConversionRules {
     [k: string]: ComponentConversionRule;
 }
 
+interface IntegrationLogo {
+    logoPath: string;
```
wwjhu, line 200:
```suggestion
    path: string;
```

test/validators/groups-validator.test.ts
```diff
@@ -60,5 +60,34 @@ describe('GroupsValidator', () => {
...
+                components: ['picture'],
+                integrationLogo: { logoPath: 123 },
+            });
+            definition.groups.push({
+                name: 'g3',
+                components: ['picture'],
+                integrationLogo: { logoPath: 'path', link: {} },
+            });
+            validator.validate();
+            expect(error).toHaveBeenCalledWith(
```
wwjhu, line 88: Missing expect for `g2` here. Or it should be a separate test

lib/components-schema-v1_11_x.ts
wwjhu: I think it is fine for now to have just bottom right. The setup allows us to easily extend it if needed.

wwjhu:
Superseeded by https://github.com/WoodWing/studio-component-set-tools/pull/179 and https://github.com/WoodWing/studio-component-set-tools/pull/180

PR #177, opened 2023-08-22
Boyscout-version-12-bump

Gecete:


wwjhu (changes requested):
Doing this needs to wait till we're about to release the feature. `test/resources/minimal-sample-1_11_x/components-definition.json` needs to contain the new features we'll add and we're not done yet.

Gecete:
Ok so, we can test it in the meantime by using `-next` tag once the other PR is merge, so if it detects that the schema version is -next it will pick the latest changes under development, but officially no new version is released yet

PR #188, opened 2023-12-11
CSH-10045: Add `chartType` property

Gecete:
As requested in this PR https://github.com/WoodWing/contentstation/pull/4809#discussion_r1422291715

In case we want to support different chartTypes or chart providers, not just **Infogram**, then we have a way to distinguish them.

### DIFF
--- a/lib/components-schema-v1_11_x.ts
+++ b/lib/components-schema-v1_11_x.ts
@@ -328,11 +328,15 @@ const componentPropertyDefinition: {
             },
             {
                 additionalProperties: false,
-                required: ['type'],
+                required: ['type', 'chartType'],
                 properties: {
                     type: {
                         enum: ['chart'],
                         description: 'Adds chart options',
+                    },
+                    chartType: {
+                        type: 'string',
+                        description: 'Chart provider',
                     },
                 },
             },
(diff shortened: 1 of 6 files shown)


lib/components-schema-v1_11_x.ts
```diff
@@ -328,12 +328,16 @@ const componentPropertyDefinition: {
...
                 additionalProperties: false,
-                required: ['type'],
+                required: ['type', 'chartType'],
                 properties: {
                     type: {
                         enum: ['chart'],
                         description: 'Adds chart options',
                     },
+                    chartType: {
+                        type: 'string',
```
wwjhu, line 338: Should this be a string or rather defined union of strings?

lib/validators/doc-chart-validator.ts
```diff
@@ -53,17 +54,27 @@ export class DocChartValidator extends Validator {
...
+        component: Component,
+        chartProperty: ComponentProperty<ComponentPropertyControlChartProperties>,
+    ) {
         if (!chartProperty.directiveKey) {
             this.error(
                 `Component "${component.name}" must configure "directiveKey" for the property with control type "chart"`,
             );
             return;
         }
+        if (!chartProperty.control.chartType) {
```
wwjhu, line 74: Checking the existence can be removed again since it is already checked through the schema. Same for the test.

wwjhu (approved):

PR #190, opened 2023-12-18
CSH-10063-Add-percentage-as-unit-in-for-text-property-fields

Gecete:
In order to customise the chart-component height, we need to be able to set a `padding-bottom` percentage, this way we can respect the aspect ratio, that is needed for any chart ([similar to the embed component](https://github.com/WoodWing/csde-components-boilerplate/blob/fd4b1b8b123737aab3f4fe546007d552fc0b60a1/component-sets/default-components/styles/_embed.scss#L8)).


Erik-Peet (approved):
Looks good to me

wwjhu (approved):

PR #195, opened 2025-03-26
ECS-608: extend icons validator to detect files that appear to be png…

wwjhu:
… but are not


lib/validators/icons-validator.ts
```diff
@@ -22,17 +22,29 @@ export class IconsValidator extends Validator {
...
+            const png = PNG.sync.read(data as Buffer);
+
+            if (png.alpha === false) {
+                this.error(`PNG icons are only supported when they are transparent`);
+            }
+        } catch (e) {
+            if (e.message === 'Invalid file signature') {
+                this.error(`Component "${name}" icon "${icon}" is not a valid PNG file`);
+            } else {
+                throw e;
```
jaapvanblaaderen, line 47: Bit in doubt if we should still throw in this case. Maybe we could also have something like this:
```ts
this.error(`Unable to validate component "${name}" icon "${icon}" is probably not a valid PNG file`);
console.error('Unexpected PNG validation error', error);
```

jaapvanblaaderen (approved):

jaapvanblaaderen (approved):
👌

PR #199, opened 2025-11-04
CSH-10704: Update to trusted publishing

wwjhu:
* Update some of the dependencies, removes eslint-plugin-prettier which slows down linting.
* Update GitHub actions to be able to publish to GitHub Packages


.github/workflows/release.yml
```diff
@@ -26,10 +31,11 @@ jobs:
 
             # https://docs.github.com/en/actions/guides/publishing-nodejs-packages#publishing-packages-to-npm-and-github-packages
             # Setup .npmrc file to publish to GitHub Packages (can replace the other setup-node action when we no longer publish to npm)
-            - uses: actions/setup-node@v4
+            - uses: actions/setup-node@v5
```
jaapvanblaaderen: Can also be `v6`?

.github/workflows/release.yml
jaapvanblaaderen: I guess specifying the token here should not be needed anymore as this is now covered by the OIDC auth? Also wondering what this action brings instead of just executing `npm publish`

jaapvanblaaderen (commented):
I have some small remarks. Maybe it's also nice to have a few words in the readme on how the OIDC auth is configured (link to: https://docs.npmjs.com/trusted-publishers)?

wwjhu:
> I have some small remarks. Maybe it's also nice to have a few words in the readme on how the OIDC auth is configured (link to: https://docs.npmjs.com/trusted-publishers)?

That makes sense, will add something

.github/workflows/release.yml
wwjhu: True, but it seems `epeli/npm-release` requires the token to be set. Not sure if we can continue using that action.

Gecete (approved):

Gecete:
the [npm settings](https://www.npmjs.com/package/@woodwing/studio-component-set-tools/access) for trusted publishing have been updated to use the publish.yml 🚀

PR #200, opened 2026-06-17
CSH-10704/trusted publishing continuation

Gecete:
# CSH-10704: Complete the switch to npm OIDC Trusted Publishing

> **Stacked on top of #199** (base branch: `update`). This PR contains the changes that finish the trusted-publishing migration.

## Context

[#199](https://github.com/WoodWing/studio-component-set-tools/pull/199) added the OIDC permission (`id-token: write`) and bumped the runner/action versions, but the actual publish step still authenticated with `NPM_TOKEN` via `epeli/npm-release`, so trusted publishing was wired up in permissions but never actually used. Nothing about npm authentication had changed yet.

This PR completes the migration: publishing now happens **token-free** over OIDC, with provenance.

## What this PR adds (on top of #199)

- **Use OIDC instead of the token.** Replace `epeli/npm-release` + `NPM_TOKEN` with a direct, token-free `npm publish`.
- **Upgrade npm to ≥ 11.5.1** (`npm install -g npm@latest`), Node 22 ships npm 10, which is below the OIDC floor.
- **Point the npm CLI at the registry** via `setup-node` `registry-url` (no `NODE_AUTH_TOKEN`).
- **Enable provenance** (`--provenance`) on both stable and prerelease publishes.
- **Merge `release.yml` + `prerelease.yml` into a single `publish.yml`.** npm allows only **one trusted publisher per package**, matched on the workflow filename, so both publish paths must share one file. Steps are gated by `github.event_name`.
- **Tighten permissions** `contents: write` → `contents: read` (the publish flow pushes nothing).

## Behavior - unchanged

| Event | Result |
|---|---|
| Push / merge to `master` | Publishes `<version>-dev.<sha>` to npm under the **`next`** tag |
| GitHub **release** published | Publishes the `package.json` version to npm (**`latest`**) **and** to GitHub Packages |

Triggers, version format, and dist-tags are identical to before, only the file structure and auth mechanism changed. GitHub Packages publishing still uses the ephemeral `GITHUB_TOKEN`.

## ⚠️ Required manual steps before publishing works

These can't live in code and must be done by a package owner (@wwjhu ??)  :

1. **Configure the trusted publisher** at `npmjs.com/package/@woodwing/studio-component-set-tools/access`:
   - Provider: **GitHub Actions** · Organization: `WoodWing` · Repository: `studio-component-set-tools`
   - Workflow filename: **`publish.yml`** · Environment: _(blank)_ · Allowed action: `npm publish`
2. **Remove the `NPM_TOKEN` repo secret** — only *after* a successful OIDC publish confirms the new path works.

## Testing

- `actionlint` passes clean on `publish.yml` and `build.yml`.
- `npm pack --dry-run` produces the expected tarball (`dist/` + `lib/`).
- Prerelease version string verified to match the previous `epeli/npm-release` format (`<version>-dev.<9-char-sha>`).
- **End-to-end OIDC** can only be validated in CI: once the trusted publisher is configured, the first merge to `master` exercises the prerelease path (publishes under `next`, doesn't touch `latest`) as a low-risk smoke test. Confirm via the green Provenance badge on npm and `npm audit signatures`.

## Reviewer notes

- The publish job is now named `publish` (was `release`/`prerelease`). These run post-merge, so they shouldn't be required status checks, worth confirming against branch protection settings.


Gecete:
@claude

.github/workflows/publish.yml
```diff
@@ -0,0 +1,81 @@
...
+            - uses: actions/setup-node@v6
+              with:
+                  node-version: 22
+                  registry-url: 'https://registry.npmjs.org'
+
+            # Trusted publishing requires npm >= 11.5.1; Node 22 still ships npm 10.
+            - name: Ensure npm supports trusted publishing
+              run: npm install -g npm@latest
+
+            - uses: bahmutov/npm-install@v1
```
danemesis, line 35: I'd stick to `npm ci`

.github/workflows/publish.yml
```diff
@@ -0,0 +1,81 @@
...
+    release:
+        types: [published]
+
+# npm allows only one trusted publisher per package, matched on the workflow
+# filename. Keeping both publish paths in this single file lets one trusted
+# publisher (publish.yml) cover stable and prerelease, fully token-free.
+permissions:
+    id-token: write # Required for OIDC trusted publishing to npm
+    contents: read
+    packages: write # Required to publish to GitHub Packages
```
wwjhu, line 17: It's publishing to npm, not github packages, right?

.github/workflows/publish.yml
wwjhu: We can move to `npm ci` but I would do that in a separate PR

.github/workflows/publish.yml
```diff
@@ -0,0 +1,81 @@
...
+            # registry-url writes an .npmrc pointing at npmjs.org. With OIDC
+            # trusted publishing, no NODE_AUTH_TOKEN/NPM_TOKEN is needed.
+            - uses: actions/setup-node@v6
+              with:
+                  node-version: 22
+                  registry-url: 'https://registry.npmjs.org'
+
+            # Trusted publishing requires npm >= 11.5.1; Node 22 still ships npm 10.
+            - name: Ensure npm supports trusted publishing
+              run: npm install -g npm@latest
```
wwjhu: I think we can move to 24 to remove the need for adding this

.github/workflows/publish.yml
Gecete: we publish to both (github packages and npm), same as in the old `release.yml`

wwjhu (approved):
👍

.github/workflows/publish.yml
danemesis: I checked prev file and see that this was carried from previous implementation 
so sure, can be done in separate PR.

danemesis (approved):

